PR #24, opened 2026-05-14
feat: support streaming responses to avoid 524 timeouts behind Cloudflare proxies

### LINKED ISSUE #23: Support streaming responses to avoid 524 timeouts behind Cloudflare-fronted LLM proxies
## Problem

When the configured `ai_base_url` is behind Cloudflare (e.g. self-hosted LiteLLM/Ollama/llama.cpp exposed via Cloudflare Tunnel), slow models reliably fail with `curl: (22) ... error: 524`. Cloudflare's edge `proxy_read_timeout` is 100s on Free/Pro/Business plans and cannot be raised. Any LLM whose non-streaming completion exceeds 100s — common for self-hosted models on large PR diffs — will 524, exhausting all `ai_primary_retries` attempts.

Example from a real run: 8/8 primary attempts returned 524, each after ~125s. Logs:

```
Primary model attempt 1/8: <model> @ <url> (openai)
curl: (22) The requested URL returned error: 524
Primary model attempt 1 failed; waiting 15s
...
```

## Why streaming fixes it

Cloudflare's 100s edge timer resets on every byte sent by the origin. A streamed response keeps the connection alive indefinitely as long as tokens flow. Time-to-first-token for self-hosted models is usually <30s even when total generation runs minutes, so streaming sidesteps 524 entirely.

## Current behavior

[`scripts/run_review.sh`](https://github.com/misospace/pr-reviewer-action/blob/v1.0.15/scripts/run_review.sh#L170-L230) sends a non-streaming request:

- `build_model_request` does not set `"stream": true` in the JSON payload for either OpenAI or Anthropic formats.
- `curl_model` uses plain `curl --data @file` with no `--no-buffer`, buffering the entire response.

## Proposed change

Add an optional input (e.g. `ai_stream` / `ai_fallback_stream`, defaulting to `true`) that:

1. Sets `"stream": true` in the request body.
2. Adds `--no-buffer` to the curl invocation.
3. Reassembles the streamed response (SSE for OpenAI `data:` lines, event-stream for Anthropic `message_delta` / `content_block_delta`) into the same shape `parse_and_validate` consumes today, before that step runs.

Defaulting to `true` improves robustness for the common case (hosted APIs all support streaming, and Cloudflare-fronted self-hosted backends need it); users on backends that don't speak SSE can opt out.

Happy to send a PR if the direction sounds right.

joryirving:
Fixed #23 

## Summary

- Add `ai_stream` input (default: `true`) that enables streaming LLM requests with `--no-buffer` curl flag
- Add `ai_fallback_stream` input for independent fallback control, defaults to `ai_stream` value
- Implement SSE response reassembly for both OpenAI and Anthropic formats into the standard non-streaming JSON consumed by `parse_and_validate`

## Motivation

When `ai_base_url` is behind Cloudflare (e.g. self-hosted LiteLLM/Ollama via Cloudflare Tunnel), slow models reliably fail with `curl: (22) error: 524`. Cloudflare's edge `proxy_read_timeout` is 100s on all plans. Streaming sidesteps this because the edge timer resets on every byte sent by the origin — as long as tokens flow, the connection stays alive indefinitely.

## Changes

- **`action.yml`**: Added `ai_stream` (default `"true"`) and `ai_fallback_stream` (defaults to `""` → resolves to `ai_stream`) inputs
- **`scripts/run_review.sh`**:
  - `build_model_request`: Adds `"stream": true` to request payload when enabled
  - `curl_model`: Adds `--no-buffer` for streaming, plus `Accept: text/event-stream` header for Anthropic
  - `reassemble_sse_response`: New function that reassembles SSE chunks (OpenAI `data:` lines or Anthropic `content_block_delta`/`message_delta` events) into the standard non-streaming JSON shape
  - Primary and fallback model calls wired to streaming when enabled

## Backwards compatibility

Users on backends that don't speak SSE can set `ai_stream: "false"` to opt out.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -59,6 +59,14 @@ inputs:
     description: Delay between primary retries in seconds
     required: false
     default: "15"
+  ai_stream:
+    description: If true, use streaming responses to avoid timeouts behind proxies with short read timeouts (e.g. Cloudflare 100s edge timer)
+    required: false
+    default: "true"
+  ai_fallback_stream:
+    description: If set, overrides ai_stream for the fallback model; defaults to ai_stream value when blank
+    required: false
+    default: ""
   allowed_source_hosts:
     description: Comma-separated list of hosts allowed for linked source fetching
     required: false
@@ -224,6 +232,8 @@ runs:
         AI_FALLBACK_API_KEY: ${{ inputs.ai_fallback_api_key }}
         AI_PRIMARY_RETRIES: ${{ inputs.ai_primary_retries }}
         AI_PRIMARY_RETRY_DELAY_SEC: ${{ inputs.ai_primary_retry_delay_sec }}
+        AI_STREAM: ${{ inputs.ai_stream }}
+        AI_FALLBACK_STREAM: ${{ inputs.ai_fallback_stream }}
         ALLOWED_SOURCE_HOSTS: ${{ inputs.allowed_source_hosts }}
         SYSTEM_PROMPT: ${{ inputs.system_prompt }}
         SYSTEM_PROMPT_FILE: ${{ inputs.system_prompt_file }}
--- a/scripts/run_review.sh
+++ b/scripts/run_review.sh
@@ -34,6 +34,8 @@ AI_FALLBACK_MODEL="${AI_FALLBACK_MODEL:-}"
 AI_FALLBACK_API_KEY="${AI_FALLBACK_API_KEY:-}"
 AI_PRIMARY_RETRIES="${AI_PRIMARY_RETRIES:-8}"
 AI_PRIMARY_RETRY_DELAY_SEC="${AI_PRIMARY_RETRY_DELAY_SEC:-15}"
+AI_STREAM="${AI_STREAM:-true}"
+AI_FALLBACK_STREAM="${AI_FALLBACK_STREAM:-$AI_STREAM}"
 ALLOWED_SOURCE_HOSTS="${ALLOWED_SOURCE_HOSTS:-github.com,api.github.com,gitlab.com,registry.terraform.io,artifacthub.io}"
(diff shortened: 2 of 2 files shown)


itsmiso-ai (approved):
## Review: Approve

**What this PR does:** Adds streaming (SSE) response support to avoid 524 Gateway Timeout errors when `ai_base_url` is behind Cloudflare or similar proxies with short read timeouts.

**Changes:**
- `action.yml`: `ai_stream` (default: `"true"`) and `ai_fallback_stream` (defaults to `ai_stream`) inputs
- `run_review.sh`: `--no-buffer` curl flag and `Accept: text/event-stream` header for Anthropic streaming; `reassemble_sse_response` function handles OpenAI `data:` lines and Anthropic `message_delta`/`content_block_delta` SSE events, reassembling them into the standard JSON shape consumed by `parse_and_validate`; primary and fallback model call sites wired correctly

**Observations:**

1. **SSE reassembly — correct.** Both OpenAI and Anthropic SSE formats are handled properly: token accumulation, content assembly, finish reason/stop_reason, and usage tracking all look right.

2. **Call site pattern works.** The chained `&&` across `curl_model`, optional `reassemble_sse_response`, and `parse_and_validate` is functionally correct — shell `&&` short-circuits on first failure, so reassembly is skipped when streaming is disabled.

3. **`ai_fallback_stream` override is clean.** `AI_FALLBACK_STREAM="${AI_FALLBACK_STREAM:-$AI_STREAM}"` gives independent fallback control without breaking the default case.

4. **Minor (not a blocker):** The jq payload always includes `stream` in the JSON body (`true` or `false`). Most backends ignore `stream: false` silently, but some may not expect it. Worth watching if any backend rejects it — a conditional jq filter to omit `stream` when `false` would be the fix.

**Recommendation: Approve.**

PR #49, opened 2026-05-19
fix: address post-review non-blocking observations for #33

### LINKED ISSUE #33: security: strip internal review metadata markers from model output
## Problem

The managed PR comment includes internal HTML metadata markers such as review fingerprint and head SHA, then appends model-generated markdown.

A malicious PR could try to influence the model into emitting fake internal markers, for example:

If later parsing scans the whole managed comment body, model output could interfere with skip/precheck behavior.

## Desired behavior

Treat internal comment metadata as reserved and strip or escape matching marker patterns from model-generated review markdown before publishing.

## Acceptance criteria

- [ ] Model review markdown cannot create or override internal metadata markers.
- [ ] Precheck reads only the trusted metadata block, or publish strips reserved markers before comment creation.
- [ ] Add regression tests for fake marker injection in review_markdown.
- [ ] Document the reserved marker behavior.

itsmiso-ai:
Addresses the two non-blocking observations from the review of #48 (issue #33):

1. **Trailing whitespace in action.yml** — removed trailing whitespace on lines 327 and 375 after the strip_metadata_markers.py calls.

2. **ai-pr-review-sha extraction in check_review_needed.sh** — added  extraction alongside  with a comment explaining both markers. The SHA is stored for future out-of-date detection but not yet used in skip logic.

CI: validate check runs are green on this branch.

Fixes #33.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,6 +14,7 @@ The action collects rich PR context (diff, files, linked issues, version hints, 
 - **`scripts/default_system_prompt.txt`** - Bundled system prompt used when no override is provided
 - **`scripts/run_evidence_providers.py`** - Runs user-defined evidence provider commands from a JSON config, parses severity/findings output
 - **`scripts/run_tool_harness.py`** - Tool harness in `plan_execute_once` mode: model plans read-only tool requests (gh_api, read_file, web_fetch, git_grep), action executes them, appends results to corpus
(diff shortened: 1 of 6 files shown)


itsmiso-ai:
Pushed fixes directly to #48 branch per LilDrunkenSmurf's request.

PR #141, opened 2026-06-07
fix: expand glob patterns in standards_file_candidates

### LINKED ISSUE #125: standards_file_candidates does not expand glob patterns
## Problem

`scripts/run_review.sh` `resolve_standards_file` (lines 126-141) iterates comma-separated candidates and checks each with `[[ -f "$candidate" ]]`:

```bash
resolve_standards_file() {
  if [[ -n "$STANDARDS_FILE" && -f "$STANDARDS_FILE" ]]; then
    return
  fi

  local candidate
  IFS=',' read -ra candidates <<< "$STANDARDS_FILE_CANDIDATES"
  for candidate in "${candidates[@]}"; do
    candidate="$(printf '%s' "$candidate" | xargs)"
    [[ -n "$candidate" ]] || continue
    if [[ -f "$candidate" ]]; then
      STANDARDS_FILE="$candidate"
      return
    fi
  done
}
```

Bash filename expansion does not happen inside `[[ -f ]]` for the literal pattern string. A candidate like `.agents/*` is checked as a literal filename, which never exists.

## Impact

Users who write `standards_file_candidates: "AGENTS.md,.agents/*"` get the second candidate silently ignored. The `.agents/*` looks like a "match any file in .agents/" pattern but is dead code.

Observed in home-ops: `standards_file_candidates: "AGENTS.md,.agents/*"` resolves to the root `AGENTS.md` only. The `.agents/*` entry is misleading documentation, not working configuration.

## Proposed fix

Expand globs inside the loop. Bash natively supports glob expansion; this is a small change:

```bash
resolve_standards_file() {
  if [[ -n "$STANDARDS_FILE" && -f "$STANDARDS_FILE" ]]; then
    return
  fi

  local candidate matches
  IFS=',' read -ra candidates <<< "$STANDARDS_FILE_CANDIDATES"
  shopt -s nullglob
  for candidate in "${candidates[@]}"; do
    candidate="$(printf '%s' "$candidate" | xargs)"
    [[ -n "$candidate" ]] || continue
    # Expand glob to a list; nullglob makes a non-matching pattern expand to nothing
    matches=( $candidate )
    for m in "${matches[@]}"; do
      if [[ -f "$m" ]]; then
        STANDARDS_FILE="$m"
        shopt -u nullglob
        return
      fi
    done
  done
  shopt -u nullglob
}
```

Notes:
- `shopt -s nullglob` makes a non-matching pattern expand to nothing — correct for "if a glob matches, use the first match; if it doesn't, skip silently".
- `shopt -u nullglob` restores the prior state, keeping the change scoped to this function.
- First-match-wins is preserved (matches the existing function behavior).

## Test/validation

Add a smoke test case: set `standards_file_candidates` to a glob like `/tmp/pr-reviewer-test/*.md` with two or more matching files. Assert the first match wins. Also test that a glob that matches no files is silently skipped (falls through to next candidate).

## Documentation

Update `AGENTS.md` (lines 66-74, conventions section) to document glob support in `standards_file_candidates`.

itsmiso-ai:
Fixes #125

## Problem

`resolve_standards_file` checks each candidate with `[[ -f "$candidate" ]]`, which does not expand glob patterns. A candidate like `.agents/*` is checked as a literal filename and never matches.

## Fix

Enable `nullglob` before the candidate loop so bash expands glob patterns to matching files. Non-matching globs expand to nothing (skipped silently). First-match-wins behavior is preserved.

- Added `shopt -s nullglob` before the loop
- Expand each candidate with `matches=( $candidate )` and iterate matches
- Restore `nullglob` off on all exit paths (`shopt -u nullglob`)


itsmiso-ai:
## Complete changes in this PR

### Fix (`scripts/run_review.sh`)
- `resolve_standards_file` now uses `shopt -s nullglob` to expand glob patterns like `.agents/*.md` instead of checking for literal filenames
- Non-matching globs expand to nothing (silently skipped via nullglob)
- First-match-wins behavior preserved

### Tests (`tests/test_standards_file_resolution.sh`) — 8 test cases, all passing:
1. Plain file candidate resolves correctly
2. Glob matches multiple files — first match wins
3. Non-matching glob silently skipped, falls through to next
4. Only non-matching globs — STANDARDS_FILE remains empty
5. Explicit `standards_file` input takes priority
6. Mixed plain and glob candidates — plain found first wins
7. Plain missing, glob found as fallback
8. Whitespace around candidates is trimmed

### Documentation (`AGENTS.md`)
- Documented glob pattern support in `standards_file_candidates`

Issue #197, opened 2026-06-10
Umbrella: agentic tool use beyond the planner loop (native tool calling, loop quality, tool surface, security)


joryirving:
Umbrella issue for agentic tool use beyond the Option A planner loop. #192 will be closed by the Option A implementation (iterated plan→execute); everything below is the future work that must NOT get lost with it. Treat this as the 1.3.x research-loop roadmap — items graduate to their own issues when picked up.

## Context

- `plan_execute_once` (shipped): one blind planning round.
- **Option A** (#192, near-term): iterated plan→execute loop — model sees round N's results before planning round N+1. Model-agnostic, no transport changes.
- **This issue**: everything beyond that.

## 1. Native tool calling (Option B) — the 1.3.x flagship

Drive the loop with real function calling (OpenAI `tools`/`tool_calls`, Anthropic `tool_use`) instead of the bespoke JSON-plan protocol. The deployment models (qwen3.6-35b, gemma-4-26b via llama.cpp/LiteLLM, `supports_function_calling: true`) are proven tool callers in opencode — model capability is NOT the blocker. Engineering blockers, each sizeable:

- **SSE tool-call delta reassembly**: `pr_reviewer/sse_reassembler.py` is text-only; OpenAI streams tool calls as incremental argument fragments across deltas, Anthropic as `input_json_delta` blocks. Needs per-format stateful reassembly + tests against real stream captures.
- **Dual-format conversation management**: multi-turn assistant messages with tool-call blocks + tool-result messages differ structurally between formats; `build_model_request`/`model_call.sh` only know single-shot system+user today.
- **Mode switching vs `ai_response_format`**: research turns need `tools`; the final verdict turn needs strict JSON (`json_object`/`json_schema`). Most backends reject both at once — the loop must drop `tools` and switch formats for the closing turn.
- **Budget enforcement on a live conversation**: stop conditions (max turns, token budget, wall clock) require truncating an in-flight conversation gracefully — "you are out of tool budget; produce the review now" as a forced final turn.
- **Fallback matrix**: per-model capability detection (or an input flag), degrading to Option A's planner loop for models that can't tool-call, and to `plan_execute_once` below that.

## 2. Loop quality upgrades (apply to A and B)

- **Per-route loop depth**: shallow/no loop on the fast route, full depth on smart/escalated routes — spend research where risk is.
- **Adaptive budget**: scale rounds/requests by PR risk classification (`pr_kind`, `risk_flags`) instead of a flat cap.
[...]

Issue #202, opened 2026-06-10
Native tool calling 2/7: multi-turn conversation/request builder


joryirving:
Part of #197 §1 (Option B stack, item 2; independent of 1/7 — can develop in parallel).

A conversation-state builder for multi-turn tool exchanges, per API format:

- OpenAI: assistant messages carrying `tool_calls`, followed by `role: tool` result messages keyed by `tool_call_id`; `tools` schema definitions for the existing executor menu (gh_api, read_file, git_grep, web_fetch, run_command).
- Anthropic: assistant content blocks with `tool_use`, user messages with `tool_result` blocks keyed by `tool_use_id`.

Likely a Python module (the bash `build_model_request` only knows single-shot system+user); unit-testable with no endpoint. Include conversation-size accounting so the loop driver (3/7) can enforce budgets.

PR #233, opened 2026-06-12
feat(sse): reassemble streamed tool calls (issue #201, 1/7 of #197 Option B)

### LINKED ISSUE #201: Native tool calling 1/7: SSE tool-call delta reassembly
Part of #197 §1 (Option B stack, item 1 — the natural starting point; no dependencies).

Extend `pr_reviewer/sse_reassembler.py` (currently text-only) to reassemble streamed tool calls for both API formats:

- **OpenAI**: `choices[].delta.tool_calls[]` — id/name arrive once, `function.arguments` accumulates as string fragments across deltas, indexed by `index`; terminated by `finish_reason: tool_calls`.
- **Anthropic**: `content_block_start` (type `tool_use`, carries id/name) + `input_json_delta` partial-JSON fragments + `content_block_stop`; `stop_reason: tool_use`.

Output: the reassembled non-streaming-equivalent response object so downstream parsing is format-agnostic. Pure parsing change, zero behavior change for existing modes.

**Tests**: fixture-driven against captured real streams from llama.cpp/LiteLLM (qwen3.6, gemma-4) and an Anthropic-format endpoint; pathological cases (interleaved text+tool blocks, split multibyte chars mid-fragment, truncated streams).

joryirving:
Fixes #201

## What

Extends `pr_reviewer/sse_reassembler.py` to reassemble streamed tool calls for both API formats, alongside the existing text-only path.

**OpenAI** — `choices[].delta.tool_calls[]` deltas indexed by `index`; `id` and `function.name` arrive once, `function.arguments` accumulates as string fragments. Reassembled into the OpenAI non-streaming-equivalent `tool_calls` shape: `[{id, type, function: {name, arguments}}]`. `arguments` is parsed back to a JSON object when the fragments form valid JSON (otherwise the raw string is preserved).

**Anthropic** — `content_block_start` (carries `id`/`name`/`type=="tool_use"`) + `input_json_delta` partial-JSON fragments + `content_block_stop`, with `stop_reason: "tool_use"`. Both are normalised to the OpenAI `tool_calls` shape on the way out, with `finish_reason` mapped `"tool_use" -> "tool_calls"` so downstream code classifies both formats identically.

Output is always the OpenAI-style dict (per the module's existing contract), so `response_parser` and downstream consumers stay format-agnostic.

## Why

Part of the Option B native-tool-calling stack in #197. Item 1/7 (SSE tool-call delta reassembly) was the natural starting point and has no dependencies.

## Backward compatibility

The existing `choices[0].message.content` string is preserved verbatim, so consumers that look only at the content field see **zero behavior change**. Verified by the full test suite (668 passed, no regressions).

## Robustness notes

- **Incomplete tool calls** (missing `id` or `name`) are dropped, not emitted half-formed.
- **Malformed partial JSON** preserves the raw string for the caller to inspect or retry.
- **Truncated streams** (no `finish_reason` / no `content_block_stop`) are flushed defensively at end-of-stream or `message_stop`.
- **LiteLLM proxy quirks** are accepted: single-dict `delta.tool_calls` (not list), `delta.input` (dict) as an alias for `partial_json` (string).
- **Plain-JSON-error fallback** at the bottom of `reassemble_sse` is now gated on *no tool calls either*, so a tool-call-only stream is not misread as a transport error.

## Tests

Adds 16 new tests in `tests/test_sse_reassembler.py`:

**OpenAI (`TestReassembleOpenAIToolCalls`):**
- `test_single_tool_call_arg_fragments` — real llama.cpp/LiteLLM shape with partial-JSON fragments split mid-key
- `test_text_then_tool_call` — interleaved prose + tool call
- `test_parallel_tool_calls_distinct_indices` — two parallel tool calls in one round
- `test_truncated_stream_flushes_partial_tool_calls` — no `finish_reason`, no `[DONE]`
- `test_incomplete_tool_call_dropped` — `id` never arrives
- `test_single_dict_tool_call_delta` — LiteLLM proxy shape
- `test_multibyte_chars_in_arguments` — non-ASCII round-trip
- `test_text_only_no_tool_calls_key` — backward-compat invariant

**Anthropic (`TestReassembleAnthropicToolCalls`):**
- `test_single_tool_use_with_input_json_delta` — basic `input_json_delta` reassembly
- `test_text_then_tool_use_then_text_interleaved` — interleaved text + tool + text
- `test_tool_use_only_no_text_content` — tool-use only, no prose
- `test_truncated_stream_defensive_flush` — no `content_block_stop` before `message_stop`
- `test_multiple_tool_use_blocks` — two `tool_use` blocks in the same response
- `test_malformed_partial_json_preserves_raw_string` — truncated partial JSON
- `test_input_json_delta_with_dict_alias` — `delta.input` (dict) shape
- `test_thinking_blocks_ignored` — `thinking` blocks must not leak

## Validation

- `pytest tests/ -v` → 668 passed
- `python3 -m py_compile pr_reviewer/sse_reassembler.py` → ok
- Module imports cleanly and the file-based helpers round-trip as expected

## Out of scope

This PR is the parsing-layer change only. It does not wire tool calls into `build_model_request`/`model_call.sh` or the conversation loop — those are items 2/7+ in the #197 stack.

### DIFF
--- a/pr_reviewer/sse_reassembler.py
+++ b/pr_reviewer/sse_reassembler.py
@@ -3,12 +3,36 @@
 Ported from the ``reassemble_sse_response`` function in ``scripts/run_review.sh``.
 Handles OpenAI chat completions streaming and Anthropic messages streaming formats,
 normalising both to a unified OpenAI-style response structure.
+
+Streaming tool calls are reassembled alongside text deltas into the OpenAI
+non-streaming-equivalent shape used by ``response_parser``/downstream callers:
+
+* OpenAI streams tool calls as ``choices[].delta.tool_calls[]`` deltas indexed
+  by ``index``; ``id`` and ``function.name`` arrive once, and
+  ``function.arguments`` accumulates as string fragments across deltas. The
+  reassembled message carries ``tool_calls=[{id, type, function: {name,
+  arguments}}]`` with ``arguments`` parsed back to a JSON object when the
+  fragments form valid JSON (otherwise the raw string is preserved).
+* Anthropic streams tool calls as ``content_block_start`` (carrying
+  ``id``/``name``/``type=="tool_use"``) plus ``input_json_delta`` partial-JSON
+  fragments plus ``content_block_stop``; ``stop_reason: "tool_use"``.
+
+The output dict therefore always uses OpenAI's ``tool_calls`` schema, regardless
+of the source format. Text-only responses are unchanged: the existing
+``choices[0].message.content`` string is preserved, so consumers that look
+only at the content field are unaffected (zero behavior change for existing
+modes).
 """
 
 from __future__ import annotations
 
 import json
 from pathlib import Path
+from typing import Any
+
+# ---------------------------------------------------------------------------
+# Public entry point
+# ---------------------------------------------------------------------------
 
 
 def reassemble_sse(response_text: str, api_format: str) -> dict:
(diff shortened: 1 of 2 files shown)


joryirving:
Reviewed the diff and probed it empirically (full suite reproduces: 668 passed). The accumulator design and defensive flushes are solid. Three findings, one of which is a contract decision worth settling **before** #202 builds on this:

### 1. `function.arguments` is a dict — diverges from the OpenAI non-streaming schema (decide now)

In OpenAI's non-streaming response shape, `function.arguments` is a **JSON-encoded string**, not an object. This PR parses it into a dict when the fragments form valid JSON:

```python
r["choices"][0]["message"]["tool_calls"][0]["function"]["arguments"]
# => dict {'a': 1}, not '{"a": 1}'
```

Two consequences:
- **Round-trip hazard for #202**: the conversation builder will echo the assistant `tool_calls` message back to the server on the next turn. Strict OpenAI-compatible servers (and llama.cpp/LiteLLM in some configs) reject a non-string `arguments`. If the builder has to re-`json.dumps` it anyway, parsing here buys nothing.
- **Inconsistent type**: dict normally, raw string when malformed — every consumer needs an `isinstance` fork.

Recommend: keep `arguments` as the accumulated **string** (spec-faithful), and let consumers parse. If the parsed form is wanted, add it under a separate key (e.g. `parsed_arguments`) so the canonical field stays typed.

### 2. Anthropic `tool_use` block without `index` is silently dropped + `finish_reason` left unnormalised

`content_block_start` with `type: tool_use` but no `index` sets `tool_block_index = None`; both the `content_block_stop` and `message_stop` flush paths require `tool_block_index is not None`, so the call never flushes. Worse, because `tool_calls` ends up empty, the `tool_use → tool_calls` finish_reason normalisation is skipped too:

```python
# repro: content_block_start without "index"
# => tool_calls: MISSING, finish_reason: "tool_use"   (downstream won't classify it)
```

The spec always sends `index`, so this needs a malformed proxy to trigger — but the fix is one line: key the defensive flushes on `current_tool_id is not None` alone (drop the `tool_block_index` check), matching the intent of the existing flush-on-new-block path.

### 3. Parallel tool calls flush in arrival order, not index order (minor)

A stream delivering `index: 1` before `index: 0` yields `[call_B, call_A]`. Dict insertion order ≠ index order. Sorting the end-of-stream flush (`for idx in sorted(tool_state)`) makes output deterministic and index-faithful.

---

None of these invalidate the approach — #2 and #3 are one-liners, and #1 is a schema decision that's cheap now and expensive after 2/7 consumes it.

joryirving:
Pushed `4d1b6de` addressing all three review findings:

1. `function.arguments` is now always the accumulated JSON-encoded **string** (OpenAI non-streaming schema) — malformed fragments pass through verbatim, consumers parse. Tests converted to a `_parsed_args` helper that enforces the string contract.
2. Anthropic flushes keyed on the open tool id instead of the block index — an index-less `tool_use` block now flushes and `finish_reason` normalises correctly (regression test added).
3. Parallel OpenAI tool calls flush in `sorted(index)` order (regression test added).

Verified: original probe cases now return `str` args / flushed block / `[call_A, call_B]`; full suite 670 passed. Auto-merge (squash) is enabled — this lands when the checks go green, and #202 can build against the string-args contract.

PR #235, opened 2026-06-12
feat(conversation): multi-turn conversation/request builder for native tool calling (2/7 of #197)

### LINKED ISSUE #202: Native tool calling 2/7: multi-turn conversation/request builder

joryirving:
Fixes #202

## What

Adds `pr_reviewer/conversation.py` — a pure-Python, I/O-free stateful builder for the OpenAI and Anthropic message shapes that the native tool-calling loop needs. The bash `build_model_request` in `scripts/model_call.sh` only knows single-shot `system+user`; this module owns the multi-turn shape so the loop driver in 3/7 can be a thin caller.

## Surface

- **`TOOL_SCHEMAS`** — JSON-Schema definitions for `gh_api`, `read_file`, `web_fetch`, `git_grep`, `run_command`. The `run_command` enum matches the executor's `ALLOWED_COMMANDS` allowlist.
- **`Conversation`** — append-only neutral event store. Mutators: `add_user`, `add_assistant_text`, `add_assistant_tool_calls`, `add_tool_result`, `add_system_note`. Introspection: `turns()`, `open_tool_call_ids()`, `approx_tokens()`. Overflow: `truncate_oldest_tool_results(max_bytes)` (UTF-8/newline-safe).
- **`to_request_payload(api_format, model, *, stream, max_tokens, temperature, verdict_turn, keep_full_history_on_verdict, response_format)`** — renders the wire body for both APIs.
- **`truncate_text`, `normalize_assistant_tool_calls_openai`** — small helpers exported for callers (e.g. the loop driver passing a reassembler-shaped `tool_calls` list straight into `add_assistant_tool_calls`).

## Verdict-turn switch (§1 "Mode switching vs ai_response_format")

`verdict_turn=True` is the closing-turn flag. Behaviour:

- **`tools` is dropped** on the wire so the strict `response_format` (`json_object` / `json_schema`) is accepted unimpeded. The OpenAI strict-mode `pr_review` schema is mirrored verbatim from `scripts/model_call.sh`.
- **History collapse default** — the accumulated conversation collapses into a single system-prompt transcript note (tool calls + result heads, "do not re-issue any tool calls" preamble). Cheapest tokens, mirrors today's single-shot contract.
- **`keep_full_history_on_verdict=True`** carries the full prior conversation through; the loop driver (3/7) can opt in per-route (e.g. smart route keeps history, fast route collapses).

Anthropic has no `response_format`; the parameter is silently accepted to keep the call sites uniform between APIs.

## #233 contract: string-args end-to-end

`function.arguments` is treated as an opaque JSON-encoded string end-to-end (OpenAI non-streaming schema):

- A string is **preserved verbatim** — malformed fragments from truncated streams round-trip, and the strict-server round-trip property holds for the assistant message echoed back on the next turn.
- A dict is serialised **once at the ingest boundary** (`add_assistant_tool_calls` / `normalize_assistant_tool_calls_openai`), then never touched again.
- The Anthropic render path is the exception (it requires an object in `tool_use.input`) and parses with a malformed-fallback to `{"_raw": ...}` so a bad fragment surfaces as data rather than disappearing.

## Tool-result batching (Anthropic)

Multiple tool results from the same executor round batch onto a single `role: user` message with multiple `tool_result` content blocks, matching Anthropic's batching convention. The OpenAI render emits one `role: tool` message per call (the OpenAI contract).

## Tests

`tests/test_conversation.py` (36 cases):

- `TestToolSchemas` — catalogue coverage and `run_command` enum
- `TestTruncateText` — newline boundary, multibyte safety (don't split a codepoint), partial-byte replacement, no-newline blob
- `TestNormalizeAssistantToolCalls` — well-formed pass-through, dict ingest at the boundary, string verbatim preservation, drop-on-missing-id/name, drop-non-dict
- `TestConversationIntrospection` — empty, turn count, open call ids, token estimate
- `TestConversationOverflow` — over-budget results shrunk, in-bounds untouched
- `TestOpenAIPayload` — full round-trip, `stream_options`, verdict-turn drop-tools + attach-`response_format`, verdict-turn keep-full-history, no-system-prompt
- `TestAnthropicPayload` — full round-trip, assistant_text block, multi-result batching, `is_error`, JSON-input parse, malformed-input fallback, `input_schema` shape, verdict-turn drop-tools
- `TestCrossFormat` — equivalent tool coverage and `json.dumps` serialisability

## Validation

- `python3 -m py_compile pr_reviewer/conversation.py` → ok
- `python3 -m pytest tests/test_conversation.py -v` → 36 passed
- `python3 -m pytest tests/` → **706 passed** (no regressions)

## Out of scope

- Wire-up into `build_model_request` / `model_call.sh` / `run_review.sh` — that lives in 3/7 (the loop driver) so this module stays endpoint-free and unit-testable.
- Tool-surface expansion (read_file line ranges, git_log, git_blame) — §3 in #197, separate work.
- Taint tracking / per-tool provenance labels — §4 in #197, ships *with* the loop driver rather than the builder.
- The `verdict_turn` transcript summariser is intentionally simple; §2 "Result summarization between rounds" in #197 is a more general upgrade that should live with the loop driver.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,7 +28,8 @@ The action collects rich PR context (diff, files, linked issues, version hints, 
 - **`metadata.py`** — Managed metadata marker (fingerprint, scope, open findings) embedded in published comments
 - **`github_context.py`** — PR metadata/linked-issue context helpers
 - **`response_parser.py`** — Tolerant model-output parsing (JSON in fences/prose, verdict + findings extraction)
-- **`sse_reassembler.py`** — Reassembles streamed SSE responses into complete bodies
+- **`sse_reassembler.py`** — Reassembles streamed SSE responses into complete bodies (including streamed tool-call deltas; `function.arguments` is the accumulated JSON string, OpenAI non-streaming shape, per #233)
+- **`conversation.py`** — Multi-turn conversation/request builder for native tool calling (#202, 2/7 of #197 Option B): append-only neutral state, OpenAI/Anthropic wire rendering, per-API tool-schema catalogue, `truncate_oldest_tool_results` budget helper, `verdict_turn` mode that drops `tools` and switches to the strict JSON `response_format`
 
 ### Publishing and output hygiene
 
@@ -66,7 +67,7 @@ run_review.sh                   → collects context → classifies → builds c
   ├─ run_evidence_providers.py  → User-defined provider commands
   ├─ run_tool_harness.py        → Tool harness planning + execution (once or loop)
   ├─ model_call.sh              → Fast/smart routing, retries, streaming, fallback
-  └─ pr_reviewer.{completeness,enforcement,escalation,carry_forward}
+  └─ pr_reviewer.{completeness,enforcement,escalation,carry_forward,conversation}
                                  → required-check validation, verdict policy, escalation, carried findings
 publish (action.yml steps)      → sanitize markdown → strip markers → build managed body → publish
   ├─ publish_mode=comment        → gh pr comment --edit-last --create-if-none (sticky)
--- a/pr_reviewer/conversation.py
+++ b/pr_reviewer/conversation.py
@@ -1,0 +1,799 @@
+"""Multi-turn conversation/request builder for native tool calling (#202).
+
+A pure-Python stateful builder for the OpenAI and Anthropic message shapes
+that the native tool-calling loop (umbrella #197 §1, item 2/7) needs. It is
+deliberately I/O-free and endpoint-free so it can be unit-tested without a
+running model server — the loop driver in 3/7 will own the actual request
+serialisation and HTTP call.
+
+Wire-shape contract (the bits callers will rely on):
+
+- OpenAI
+    * assistant turn: ``{"role": "assistant", "content": str|None,
+      "tool_calls": [{"id", "type": "function", "function": {"name",
+      "arguments"}}]}``
+    * tool result turn: ``{"role": "tool", "tool_call_id": <id>,
+      "content": str}``
+    * top-level ``tools``: list of ``{"type": "function",
+      "function": {"name", "description", "parameters": JSON-Schema}}``
+
+- Anthropic
+    * assistant turn: ``{"role": "assistant", "content":
+      [{"type": "text", "text": ...}, {"type": "tool_use", "id",
+      "name", "input"}]}`` — text and tool_use blocks may interleave.
+    * tool result turn: ``{"role": "user", "content":
+      [{"type": "tool_result", "tool_use_id": <id>, "content": str|list,
+      "is_error": bool?}]}``
+    * top-level ``tools``: list of ``{"name", "description", "input_schema":
+      JSON-Schema}``
+
+The verdict-turn contract: ``ai_response_format`` only applies to the
+closing turn (it forces a strict JSON verdict). For that turn, ``tools`` is
+omitted and the accumulated conversation can either be carried through in
+full (default off, expensive) or collapsed into a single system-prompt
+transcript note (default on, mirrors today's single-shot behaviour). See
+``Conversation.to_request_payload`` for the flag.
+
+The budget helpers in this module (rough token estimate + graceful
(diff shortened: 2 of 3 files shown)


joryirving:
Reviewed with empirical probes against the rendering layer. The architecture is right — neutral append-only events with wire rendering at emission, the #233 string-`arguments` contract is consumed correctly (verbatim strings, serialise-once-at-boundary for dicts), `TOOL_SCHEMAS` verified in lockstep with the executor's arg names, and the budget helpers are sensible. Three findings, all reproduced (the test suite passes but never asserts these paths):

### 1. 🔴 Verdict turn keeps `tools` when `response_format` is off

```python
p = conv.to_request_payload("openai", "m", verdict_turn=True, response_format=None)
# "tools" in p → True
```

The module's own contract says the closing turn drops `tools` unconditionally; `_to_openai_payload` only drops them when `response_format` is `json_object`/`json_schema` (the `if verdict_turn and response_format in (...)` / `else` structure couples the two decisions). Masked in our fleet (everything sets `json_object`) but broken for `ai_response_format: off`. The Anthropic side does it right (`if not verdict_turn`).

### 2. 🔴 Collapsed verdict turn emits an empty/degenerate `messages` array

With `keep_full_history_on_verdict=False` (the default), `messages = []` wipes everything — including any closing user instruction the driver appended:

- **Anthropic: hard 400** (`messages` requires at least one message, first must be `user`).
- OpenAI: a system-only conversation — accepted but degenerate.

The collapsed turn needs one closing `user` message carrying the "produce the final JSON verdict now" instruction.

### 3. 🟡 The reassembler's output shape is silently dropped

`add_assistant_tool_calls` expects flat `{id, name, arguments}`; feeding it `sse_reassembler`'s nested `{id, type, function: {name, arguments}}` records **zero calls** (the natural pipeline reassembler → conversation loses every tool call unless the driver flattens manually). The tolerant nested-shape normaliser already exists in this module — `add_assistant_tool_calls` should accept both shapes.

### Minor
- `truncate_text`'s comment says cap-of-5 on `"a\nb\nc\nd\ne"` yields `"a\nb\nc"`; the code yields `"a\nb"` (comment wrong, code fine).
- `gh_api` schema: `required: ["endpoint"]` while advertising the `path` alias property means a model using only `path` emits schema-invalid args that the executor would actually accept. Harmless today (tool schemas aren't strict-enforced), worth a note.

Pushing fixes + regression tests directly to this branch (same treatment as #233/#234).

joryirving:
Pushed `d8c2e54` addressing all three findings:

1. **Verdict turn drops `tools` unconditionally** — decoupled from `response_format`; `ai_response_format: off` no longer leaves the tool catalogue on the closing call.
2. **Collapsed verdict turn keeps one closing user instruction** (`VERDICT_USER_INSTRUCTION`) — no more empty `messages` array (Anthropic 400) or system-only conversation (OpenAI degenerate).
3. **`add_assistant_tool_calls` accepts the reassembler's nested shape** alongside the flat form — the reassembler → conversation pipeline no longer silently drops calls.

Re-probed all three (tools absent with rf=off; Anthropic verdict messages `['user']`; nested call recorded) and added regression suites; the two tests that encoded the empty-messages behavior as intent were updated to the corrected contract. Full suite: **712 passed**. Auto-merge (squash) armed — 3/7's loop driver can now build directly against `Conversation` + `sse_reassembler` with no manual flattening.

Issue #258, opened 2026-06-14
Tech debt / 2.0 consolidation audit (after #197 lands)


joryirving:
Tracking epic for a deliberate audit/refactor pass to reduce sprawl, **scheduled after the #197 native-tool-calling umbrella ships as 1.3.0** — not a feature release. Because the high-value cleanups are interface-affecting, this is a natural **2.0.0**.

This is a living inventory: append items as we trip over them during the remaining feature work. Each should note the smell and the proposed consolidation.

## Sequencing decision
1. Finish #197 features as planned (#204 streaming, #205, #206 security gate, #245 MCP) + Forgejo #223/#224/#226/#228 → **ship 1.3.0** to close the umbrella.
2. *Next* cycle = this audit, **not** 1.4.x features.
3. Interface-affecting cleanups (input-surface pruning, publish-step restructure, dropping deprecated modes) → cut as **2.0.0**, semver-honest.

## Seeded inventory
- [ ] **`action.yml` duplicate publish-step `env:` blocks** (~910-line manifest). Three publish steps (`comment` / `review_comment` / `review_verdict`) copy-paste near-identical `env:` mappings that drift independently — this is exactly what produced the duplicate-`PLATFORM` bug that broke v1.2.10 (#255). DRY this (shared env source / fewer publish steps / generation). Highest-value target.
- [ ] **Collapse tool modes.** `plan_execute_once`, `plan_execute_loop`, and `native_loop` coexist with two planning paths. Once `native_loop` is proven on the founding use case, deprecate/remove the plan_execute paths — large code reduction. (Interface change → 2.0.)
- [ ] **Split `run_tool_harness.py`.** It currently holds transport (`run_chat_request`), the read-only tool executors, the legacy planner, and the native loop in one large file. Separate transport / executors / drivers.
- [ ] **Audit the input surface.** Many `tool_*`, `evidence_*`, `escalate_*`, `forgejo`, and platform inputs. Group, document, and prune redundant ones (some may be obsolete post-mode-collapse). (Interface change → 2.0.)
- [ ] **Dual-backend (GitHub + Forgejo) consolidation** — `platform_api.sh`, `publish_helpers.sh`, `forgejo_backend.py`. Added fast; field-shape bugs recurred in review. Pair with #252 (golden fixtures / real-backend validation) so the abstraction is tested against real shapes.
[...]

PR #261, opened 2026-06-14
test(security): exfiltration red-team suite for the native loop (#206)

### LINKED ISSUE #206: Native tool calling 6/7: security pass + red-team fixtures
Part of #197 §4 (gates the 1.3.0 release — must land before native_loop defaults on anywhere).

Multi-turn means fetched content can steer subsequent tool calls. Deliverables:

- Taint posture: tool results fenced as data with per-tool provenance labels in the conversation; verify the model-visible framing matches plan_execute_loop's UNTRUSTED DATA discipline.
- Allowlist review for looped runs (hosts, gh_api repos, commands) — decide whether loop mode gets a stricter default profile.
- Exfiltration analysis: hostile PR + loop inducing reads of sensitive-but-allowlisted content into a published review (the workspace contains the PR's own hostile code).
- Red-team test suite: hostile fixtures (PR body, fetched pages, file contents carrying injection) asserting the loop does not follow embedded instructions. Same bar as the #183 audit items.

joryirving:
Part of #206 (6/7 of #197, gates the 1.3.0 release). Reclaimed after the Saffron `agent/saffron-escalated` claim went stale (label applied ~5.7h ago, no PR — the only linked PR #250 was the already-merged taint posture).

Closes the **exfiltration analysis** + **red-team test suite** deliverables; the **taint-result fencing** landed earlier in #250.

## What's here
A model-independent red-team suite (`tests/test_native_loop_exfil_redteam.py`, 23 cases) that documents the loop's exfiltration threat model and asserts every executor control holds against a hostile PR — driving `execute_tool_request` directly so it fails loudly if a future change weakens the surface:

| Vector | Control asserted |
|---|---|
| Read a secret file | `read_file` `SENSITIVE_PATH_RE` (.env/.pem/.key/credentials/secrets/id_rsa) |
| Escape the workspace | `read_file` workspace containment (traversal + absolute paths) |
| Leak a secret in an allowed file | `mask_secrets` on output (e.g. `ghp_…`) |
| Exfiltrate to an attacker host | `web_fetch` host allowlist, incl. `user@host` + suffix confusion |
| Smuggle a host into search | `web_search` query URL-encoded into the fixed operator `search_url` |
| Reach sensitive GH endpoints | `gh_api` repo + read-only-prefix allowlist + secrets/environments/dispatches deny |
| Run arbitrary shell | `run_command` named-argv allowlist; raw shell rejected |

## Allowlist review (for sign-off — the remaining #206 decision)
Reviewed the executor surface for whether `native_loop` should get a **stricter default allowlist profile**. **Recommendation: no separate profile.** Rationale:
- The executor is already tightly constrained and **read-only**: workspace-scoped reads with sensitive-path blocking, exact-match host allowlist, repo+prefix-allowlisted gh_api with sensitive-endpoint denies, and `run_command` restricted to named argv definitions (no model-controlled shell).
- It is **identical across all tool_modes**, and `plan_execute_loop` *already* exercises multi-round dynamics with the same executor — so `native_loop` adds no new per-call capability, only more turns through the same gate.
- A divergent loop-only profile would add surface/branching that cuts against the consolidation goal (#258) for marginal benefit.

If you'd rather lock loop mode down further (e.g. drop `run_command` or restrict `gh_api` to the current repo by default), say so and I'll add it as a follow-up — otherwise this + #250 closes #206.

829 tests pass on Python 3.14.

### DIFF
--- a/scripts/run_tool_harness.py
+++ b/scripts/run_tool_harness.py
@@ -322,13 +322,25 @@ def fetch_url(url, allowed_hosts, request_timeout=25):
 
 def read_file(path, workspace_root):
     """Read a file with path-traversal protection."""
-    resolved = Path(workspace_root) / path
+    # Reject embedded null bytes before touching the filesystem: pathlib raises
+    # ValueError (not OSError) on them, and a NUL can truncate the path at the C
+    # layer of an underlying syscall, so an early explicit reject is safest.
+    if "\x00" in path:
+        return {"error": "Null byte in path"}
+
+    root = Path(workspace_root).resolve()
     try:
-        resolved = resolved.resolve()
-    except OSError:
+        # resolve() also collapses symlinks, so a symlink that lives inside the
+        # workspace but points outside it is normalised to its real target and
+        # caught by the containment check below.
+        resolved = (root / path).resolve()
+    except (OSError, ValueError):
         return {"error": f"Cannot resolve path: {path}"}
 
-    if not str(resolved).startswith(str(Path(workspace_root).resolve())):
+    # Containment via is_relative_to, NOT str.startswith: startswith wrongly
+    # accepts a sibling directory whose name shares the workspace as a prefix
+    # (e.g. resolving to /work/repo2 passes a /work/repo prefix test).
+    if not resolved.is_relative_to(root):
         return {"error": "Path escapes workspace root"}
 
     if SENSITIVE_PATH_RE.search(resolved.name):
--- a/tests/test_native_loop_exfil_redteam.py
+++ b/tests/test_native_loop_exfil_redteam.py
@@ -46,12 +46,12 @@ _REPO = "owner/repo"
 _HOSTS = ["github.com", "docs.siderolabs.com"]
 
 
-def _exec(tool, args, tmp_path, *, allowed_repos=None, hosts=None, search_url=""):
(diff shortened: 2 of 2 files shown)


joryirving:
Addressed the review's ⚠️ edge-case finding (null bytes / symlinks) — and it turned up two **real bugs** in `read_file`, not just test gaps:

1. **Containment-prefix bypass:** the guard used `str(resolved).startswith(str(root))`, which admits a sibling dir sharing the workspace's name prefix (resolving to `…/repo2/secret` passes a `…/repo` test). Switched to `Path.is_relative_to`, which respects path boundaries. Verified the old check returned `True` for `/tmp/ws2` vs `/tmp/ws`.
2. **Null bytes:** an embedded NUL makes pathlib raise `ValueError` (not `OSError`), so it slipped past `except OSError` onto the broad outer catch. Now rejected explicitly up front, and `resolve()` catches `ValueError`.

Symlink escape was already defended (`resolve()` collapses the link → containment rejects an out-of-workspace target) — now locked in by a test.

+5 red-team cases (null bytes ×3, symlink escape, sibling-prefix escape); 28 in the suite, 836 total on 3.14. So this PR now closes the implementation gap the review flagged, not just the test gap.

Issue #207, opened 2026-06-10
Native tool calling 7/7: A/B eval harness (decision gate for 1.3.0)


joryirving:
Part of #197 §5 (parallel-track — no dependencies; worth doing EARLY since it gates the rest).

Benchmark harness comparing the same PR corpus reviewed three ways: tools off / plan_execute_loop / native_loop — measuring findings quality (vs known-good findings), token usage, and wall clock. Seed the benchmark set from past PRs with verified findings (the #110/#183 audit PRs are natural candidates).

Decision gate recorded in #197: native tool calling proceeds past 3/7 only if it beats the planner loop measurably. 1.3.0 ships when the stack + 6/7 land and this harness says it is worth it.

Issue #203, opened 2026-06-10
Native tool calling 3/7: loop driver (new tool_mode, non-streaming first)


joryirving:
Part of #197 §1 (Option B stack, item 3 — depends on 2/7; streaming arrives in 4/7).

New `tool_mode` (e.g. `native_loop`) driving an agentic exchange: send corpus + tools → execute returned tool calls via the EXISTING executor (allowlists/caps/timeouts unchanged) → append results → repeat until the model stops calling tools or budgets hit. Non-streaming requests only in this issue.

- Budgets: total tool calls (reuse `tool_max_requests`), rounds (`tool_max_rounds`), wall clock.
- Capability fallback: a model that never emits tool calls degrades to the plan_execute_loop planner path, then plan_execute_once.
- Results land in the corpus exactly like today's Tool Harness Findings section.

**Tests**: mock-server conversation scripts (tool call → result → second call → stop), budget exhaustion, degradation paths.

joryirving:
## Acceptance criterion: the founding use case (no recipes)

The loop driver isn't done when the loop mechanically works — it's done when it's **smart** on the PR shape this action was originally built for:

> Given a renovate Kubernetes version bump in a Talos-managed repo (canonical live fixture: [home-ops#7462](https://github.com/joryirving/home-ops/pull/7462), k8s `v1.36.1 → v1.36.2` touching `talos/*/machineconfig.yaml.j2`), the reviewing model — with **no Talos-specific prompting, steering, or evidence provider** — must on its own:
> 1. notice the bumped artifact runs on a host platform visible in the repo (the kubelet image sits in a Talos machineconfig; the installer tag in the same file carries the deployed Talos version, e.g. `v1.13.4`),
> 2. chain the tools: `read_file` → extract the Talos version → `web_fetch` the versioned support matrix (`talos.dev/v1.13/introduction/support-matrix/`, host already allowlisted) — hop N+1 depending on hop N's result,
> 3. **cite the matrix in the published review** as evidence for the verdict, instead of the current pattern-reasoned "patch release, low risk" rubber stamp.

[home-ops#7462 is being left open](https://github.com/joryirving/home-ops/pull/7462) as the live fixture; it should also be encoded as a scored scenario in the #207 eval harness so this becomes the standing regression test for "is the loop actually agentic," not a one-off demo.

[...]

PR #262, opened 2026-06-14
feat(tool-loop): in-conversation verdict for native_loop (5/7 of #197)

### LINKED ISSUE #205: Native tool calling 5/7: final-turn mode switch (drop tools, strict JSON)
Part of #197 §1 (Option B stack, item 5 — depends on 3/7).

The research turns need `tools`; the final verdict turn needs strict JSON (`ai_response_format` json_object/json_schema) — most backends reject both at once. Implement the closing turn: drop tool definitions, switch response format, and force it when budgets expire ("you are out of tool budget; produce the review now"). This is also where verdict/findings parsing reconnects to the existing pipeline (verdict_policy, carry-forward, enforcement — all unchanged).

joryirving:
Closes #205. **Option 1** (your call): the native loop's final turn produces the review verdict itself, preserving the multi-hop reasoning trajectory, instead of flattening evidence into the corpus for a separate review call.

## Why
This is the lever for the founding use case. Today the loop gathers evidence, the results get folded into the corpus, and a **separate** review call re-derives the verdict — throwing away the reasoning trajectory that connected the hops (read machineconfig → search → fetch matrix → reason). Option 1 keeps the verdict **in-conversation**, so the model reasons forward from its own chain.

## How
- **`run_native_loop`**: after a non-degraded loop, swap `NATIVE_LOOP_SYSTEM` (an evidence-gatherer prompt — "reply with a plain-text summary") for the **reviewer system prompt**, re-inject the **full review corpus** the loop never saw (it runs on the compact planning context, missing the full diff), **drop tools**, and force a strict-JSON verdict via the #202 mechanism (`verdict_turn=True, keep_full_history_on_verdict=True` — first wired here). Streams (on #259) so the long verdict generation doesn't 524. Writes the response to `ai-response.primary.json` and sets `native_loop_verdict_produced`.
- **`run_review.sh`**: when the loop produced a verdict, parse it and **skip the separate review call**. A degraded/oversized/garbled verdict simply fails to parse → falls back to the standard corpus review. **So this only ever adds capability.** Downstream (`verdict_policy`, carry-forward, completeness, enforcement) is unchanged.
- **OpenAI only**: an Anthropic verdict turn after trailing `tool_result` (user-role) blocks would create adjacent user turns (a 400). `native_loop` runs on the OpenAI primary in practice; Anthropic keeps the standard corpus review.

## Corpus completeness (the subtlety I flagged)
The loop only ever saw the compact planning context, so a naive verdict would have **less** context than today (no full diff). The verdict turn re-injects the full corpus (`review-corpus.truncated.md`, the pre-harness build — full diff/standards, tool results already in-conversation), so net context ≈ today's review call + the reasoning deltas.

## Tests / validation
Verdict turn drops tools + forces JSON + re-injects the corpus + writes the primary response; Anthropic path is skipped; smoke (25) + 838 suite green on Python 3.14. **The real validation is the #207 eval on home-ops#7462** (does in-conversation verdict move the capability pass rate?) — to run on the GPU endpoint when you're ready.

### DIFF
--- a/scripts/run_review.sh
+++ b/scripts/run_review.sh
@@ -1304,6 +1304,26 @@ The corpus lists Open Findings From the Previous Review. Answer EVERY one: inclu
   log "Carry-forward active: $(jq 'length' previous-findings.json) open finding(s) from the previous review"
 fi
 
+PRIMARY_OK=0
+# native_loop in-conversation verdict (#205): when the tool loop produced its
+# own verdict (final turn, full reasoning history preserved), it wrote the
+# response to ai-response.primary.json. Parse it and skip the separate review
+# call. A parse failure (degraded/oversized/garbled verdict) falls through to
+# the standard corpus review below, so this only ever adds capability.
+NATIVE_VERDICT_USED=0
+if [[ "$(printf '%s' "$TOOL_MODE" | tr '[:upper:]' '[:lower:]')" == "native_loop" ]] \
+  && [[ "$(jq -r '.native_loop_verdict_produced // false' tool-harness.json 2>/dev/null)" == "true" ]] \
+  && [ -s ai-response.primary.json ]; then
+  log "native_loop produced an in-conversation verdict; using it and skipping the separate review call"
+  if parse_and_validate ai-response.primary.json; then
+    PRIMARY_OK=1
+    NATIVE_VERDICT_USED=1
+  else
+    log "native_loop verdict did not parse; falling back to the standard review call"
+  fi
+fi
+
+if [ "$NATIVE_VERDICT_USED" -ne 1 ]; then
 build_model_request \
   "$AI_API_FORMAT" \
   "$AI_MODEL" \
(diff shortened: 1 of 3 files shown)


joryirving:
Added a second commit folding in the last documented native_loop production-parity gap (from the #203 work): **`AI_TOKENS_PARAM` support**.

The conversation builder always emitted `max_tokens`, but the bash review path honors `AI_TOKENS_PARAM` (`max_tokens`|`max_completion_tokens`) — newer OpenAI models **reject** `max_tokens` and require `max_completion_tokens`. So `native_loop` was silently broken for those models while the standard review call worked. Now `to_request_payload` (OpenAI only; Anthropic has no such field), `drive_tool_loop`, and `run_native_loop` (loop turns **and** the #205 verdict turn) all honor it. Belongs here because the wiring touches the verdict turn this PR adds.

843 tests pass on 3.14. With this, native tool calling (#197 §1) is at production parity with the standard review path.

Issue #263, opened 2026-06-14
Native-loop prompt caching: cache-aware payload construction (per-format)


joryirving:
The native tool-calling loop (#197 §1) is an agent harness, so prompt-cache reuse should be a **deliberate, first-class** property of how it builds payloads — not left to accident. Caching is per-format, so "supporting it" means different things per backend.

## How caching works per format
- **OpenAI / llama.cpp (the self-hosted `review` deployment):** **automatic** — the server reuses the longest stable *token prefix*; no markers. Supporting it = keep the prefix **stable and append-only** (one system prompt for the whole review, stable tool ordering, corpus up front, only tool results growing at the tail).
- **Anthropic (smart tier):** **opt-in** — caches nothing unless the request carries `cache_control: {type: "ephemeral"}` breakpoints (max 4), caching the prefix up to each marker. Our `_to_anthropic_payload` sets none today, so Anthropic never caches.

## Current state
- ✅ **Loop turns are cached** on OpenAI/llama.cpp: the conversation grows append-only with a stable system+tools prefix, so each multi-hop turn only prefills the new delta.
- ❌ **The verdict turn (#205) busts the cache.** Two divergences near the front of the prompt:
  1. It **swaps the system prompt** (`NATIVE_LOOP_SYSTEM` → reviewer prompt). The system is token 0, so this invalidates the entire cached prefix — the verdict turn (the largest prefill: full conversation + re-injected corpus) re-runs from scratch.
  2. It **drops `tools`** (required so the backend honors strict JSON). Tool schemas render near the top, so removing them invalidates from that point onward too.
- 📊 Eval data (N=3, home-ops#7462, native_loop, `AI_STREAM=true`, swap in place): 101–169s/run; the verdict turn re-prefill is part of that.

## Plan
1. **Stable unified system prompt (no swap).** Build `system = <reviewer prompt> + <tool-use preamble>` once, used for the whole review. Removes the token-0 cache-bust, lets the verdict turn reuse the system block, and gives the model reviewer context *during* gathering (better grounding). *(unit-testable, no endpoint)*
2. **Anthropic `cache_control` breakpoints** on the stable prefix blocks (system, tools) in `_to_anthropic_payload`. *(format completeness; lower deployment relevance since native_loop runs on the OpenAI primary)*
[...]

PR #264, opened 2026-06-14
fix(tool-loop): prompt-cache stable system (#263) + restore AI_TOKENS_PARAM parity

joryirving:
Two native-loop production-quality follow-ups. **Do not merge before the capability re-eval** (commit 1 changes the loop's system-prompt content).

## 1. Stable unified system prompt for prompt-cache reuse (#263, Part 1)
The #205 verdict turn swapped the system prompt (`NATIVE_LOOP_SYSTEM` → reviewer prompt), invalidating llama.cpp/OpenAI's cached prefix at **token 0** — so the verdict turn (the largest prefill) re-ran from scratch. Now there's **one stable system for the whole review** = reviewer prompt + a tool-use preamble, used for both the loop and the verdict turn. No mid-conversation swap → the cached prefix survives across every turn, and the model gathers evidence already knowing what it's reviewing for. Falls back to the tool-only system (verdict skipped) when no reviewer prompt resolves.

New test asserts the system is identical across the loop turns and the verdict turn. **Remaining in #263:** Part 2 (Anthropic `cache_control` breakpoints) and Part 3 (keep `tools` on the verdict turn to preserve the corpus+results prefix — the big win, needs endpoint validation).

## 2. Restore AI_TOKENS_PARAM parity (was dropped from #262's merge)
Heads up: the `max_completion_tokens` parity fix was a second commit on the #205 branch but **didn't make it into #262's squash** — `main` has no `tokens_param`. Re-applied here: the builder always emitted `max_tokens`, but the bash path honors `AI_TOKENS_PARAM` and newer OpenAI models reject `max_tokens`, so `native_loop` is silently broken for them. `to_request_payload`/`drive_tool_loop`/`run_native_loop` now honor it (loop + verdict turns); Anthropic unchanged (no such field).

## Validation
844 tests + smoke (25) green on Python 3.14. The **#207 capability re-eval on home-ops#7462** is the gate for commit 1 (does the unified system regress capability? does the verdict turn get faster?) — to run on the GPU endpoint.

Closes part of #263. Relates to #197.

### DIFF
--- a/pr_reviewer/conversation.py
+++ b/pr_reviewer/conversation.py
@@ -709,6 +709,7 @@ class Conversation:
         verdict_turn: bool = False,
         keep_full_history_on_verdict: bool = False,
         response_format: str | None = None,
+        tokens_param: str = "max_tokens",
     ) -> dict[str, Any]:
         """Render the conversation as a wire-ready request body.
 
@@ -738,6 +739,7 @@ class Conversation:
             verdict_turn=verdict_turn,
             keep_full_history_on_verdict=keep_full_history_on_verdict,
             response_format=response_format,
+            tokens_param=tokens_param,
         )
 
     def _to_openai_payload(
@@ -750,6 +752,7 @@ class Conversation:
         verdict_turn: bool,
         keep_full_history_on_verdict: bool,
         response_format: str | None,
+        tokens_param: str = "max_tokens",
     ) -> dict[str, Any]:
         system = self.system
         messages = self._render_openai_messages()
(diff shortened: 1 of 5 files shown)


joryirving:
### Capability re-eval (N=3, home-ops#7462, cache-busted) vs the system-swap baseline

| | capability | read_machineconfig | consulted_matrix | cited | wall avg |
|---|---|---|---|---|---|
| baseline (swap, main) | 0/3 | **0/3** | 0/3 | 3/3 | **134s** |
| #264 (unified system) | 0/3 | **2/3** | 0/3 | 3/3 | **203s** |

**Two findings:**
1. ✅ **Grounding improved, no regression.** Every #264 run now leads with `read_file` and reads the deployed machineconfig first in 2/3 (was 0/3) — the unified reviewer+tools system gives the model review context *while gathering*, so it version-grounds before searching. That's a real step toward the V3 behavior, achieved structurally. Strict capability is still 0/3 (it reads config + searches but never reaches the actual support-matrix URL — still a steering gap for the V3 prompt).
2. ❌ **Not faster — ~50% slower.** Part 1's cache saving (verdict-turn system block) is small and dominated by (a) the larger unified system prompt and (b) the model doing *more* gathering now. For similar tool-call counts (8 calls: 252s vs 169s), #264 is slower — which **hints the llama.cpp prefix cache may not be hitting as hoped, but I can't confirm without telemetry.**

**Implication:** the speed goal needs **cache-hit/token telemetry first (measure before optimize)**, then Part 3 (keep tools on the verdict turn). #264 stands on its own as a **grounding/correctness win** (safe to merge), but it is *not* the latency win — that's gated on telemetry + Part 3.

Issue #265, opened 2026-06-14
1.3.0 release gate: complete + fully-test all first-class harness capabilities


joryirving:
Decision (2026-06-14): **1.3.0 ships only when the native-loop harness is feature-complete and fully tested** — no incremental "ship the machinery, add the rest later." Motivated by the v1.2.10 busted release; we want the harness done right before cutting.

§1 native tool calling is already complete (#197 §1). This tracks the remaining first-class harness capabilities (the #263 catalog) as the 1.3.0 gate.

### Prompt caching (#263)
- [x] Part 1 — stable unified system prompt (no verdict-turn swap) — PR #264
- [ ] Part 2 — Anthropic `cache_control` breakpoints on the stable prefix
- [ ] Part 3 — keep `tools` on the verdict turn to preserve the prefix *(needs GPU eval: does the model still emit clean JSON with tools present?)*

### Tool surface (#197 §3)
- [ ] `read_file` line ranges (big files blow the response cap)
- [ ] `git_log` / `git_blame` (read-only history)
- [ ] diff-context expansion (N lines around a hunk)

### Loop quality (#197 §2)
- [ ] Per-route loop depth (shallow fast route / full smart route)
- [ ] Adaptive budget by risk (`pr_kind` / `risk_flags`)
- [ ] Parallel tool execution within a round
- [ ] Result summarization between rounds *(needs a summarization model call — GPU)*
- [ ] Cross-run evidence memory (carry gathered evidence across incremental reviews)

### Observability
- [ ] Per-run token/cost + cache-hit telemetry in the harness output

### MCP (#245)
- [ ] Read-only, allowlisted, fork-gated MCP tools in the loop (`tool_mcp_servers`)

## Also restored
- [x] `AI_TOKENS_PARAM`/`max_completion_tokens` parity (was dropped from #262's squash-merge) — PR #264

## Testing bar
- Pure-code capabilities: comprehensive unit tests, green on Python 3.14 (CI now pins 3.14 + dup-key lint, #256).
- Behavior/quality capabilities (caching Part 3, summarization, cross-run memory, MCP): unit-tested + validated on the GPU endpoint via the #207 eval before merge.
- No capability regression on home-ops#7462; verdict-turn wall-clock should improve with caching.

Relates to #197, #263, #245.

joryirving:
Progress (overnight):
[...]

Issue #193, opened 2026-06-10
Incremental reviews: cumulative verdict via carried-forward findings (one-push blocker amnesia) + header cosmetic


joryirving:
### 1. Cosmetic: move the incremental tag into the header

Incremental reviews currently render as:

> # AI Automated Review
> _Incremental review: reviewed changes since the last managed review. ..._
> _Analysis engine: review@https://litellm.jory.dev/v1 (openai) (incremental delta review)_

The "(incremental delta review)" suffix belongs on the primary header, not buried on the engine line: `# AI Automated Review (incremental)`. Drop the engine-line suffix.

### 2. Substantive: incremental reviews have one-push blocker amnesia

Scenario: full review finds blockers A, B, C (request_changes). Author pushes a commit fixing only A.

**What happens today** (verified against `check_review_needed.sh` scope resolution and the `review_verdict` publish step):

1. The push gets an **incremental** review of just the delta. The corpus contains the delta diff but **not the previous review's findings** — the model has no idea B and C exist, so it plausibly says "approve".
2. The publish guardrail catches this once: incremental + `baseline_clean=false` → native APPROVE is blocked, REQUEST_CHANGES submitted instead. Good.
3. **But** the metadata marker records `review_result` from the *model verdict* (approve → `clean`), not the published event. The **next** push sees a clean baseline → approval unblocked. Unfixed blockers survive exactly one push of scrutiny.
4. Compounding: the `verdict` action output is `approve` immediately (only the native review event is downgraded), and since v1.2.1/v1.2.2 the old reviews carrying B and C are dismissed and minimized — the carried state is the only memory left.
5. Minor: the blocked-approval explanation appended to the body says "requires `allow_approve: true`", which is wrong for the dirty-baseline path — the actual reason is the unclean baseline.

## Proposed fix: carry forward unresolved findings (cumulative verdict, still delta-priced)

Making the verdict cumulative does **not** eliminate the point of deltas — the carried state is a few KB of structured findings, not a re-review of the whole PR:

1. **Persist findings in the marker.** The metadata marker (already JSON, built by `build_metadata_marker`) additionally carries the review's `findings[]` — capped (e.g. 20 findings, messages truncated) and only when `review_result=issues`.
[...]

PR #272, opened 2026-06-15
feat(tool-loop): cross-run evidence memory across incremental reviews (#197 §2)

joryirving:
## What

A `native_loop` review gathers evidence with read-only tools (read a config, fetch a support matrix, grep callers), then **throws it away** — so the next incremental review of the same PR re-gathers it from scratch, burning tool budget and latency on evidence that didn't change.

This carries that evidence across runs via the **same channel as carried-forward findings (#193)**: the JSON metadata marker in the action's own review comment. It's a direct sibling of `pr_reviewer/carry_forward.py`.

This is the **last #265 harness capability**.

## How (mirrors carry_forward end-to-end)

- **`pr_reviewer/evidence_memory.py`** (new):
  - `build_evidence_digest` — prefer the loop's own closing summary; else a **deterministic ledger** of `tool args → result snippet`, capped. No extra model call.
  - `load_evidence_memory` — re-sanitize on read (the digest came from an attacker-influencable comment).
  - `render_evidence_memory_section` — corpus section with **fail-safe framing**: prior evidence is *context, not ground truth* — re-verify anything the delta touched; treat as untrusted data.
- **`run_tool_harness.py`** — after a native_loop review, write `result["evidence_digest"]` (from `final_text` or the executed calls) into `tool-harness.json`.
- **`run_review.sh`** — emit the `evidence_digest` step output; inject the prior-evidence section into the incremental corpus alongside carried findings.
- **`publish_helpers.sh`** — `build_metadata_marker` persists `evidence_digest` (tagged with `head_sha`, capped), gated by `TOOL_EVIDENCE_MEMORY`.
- **`check_review_needed.sh`** — precheck parses it → `previous-evidence.json` (written to a file, **never eval'd**; control/markup chars stripped).
- **`action.yml`** — `tool_evidence_memory` input (default `"true"`), wired into the review step + all three publish steps. README documents it.

## Decisions

- **Channel: comment marker** — no new permissions, forge-agnostic, consistent with the rest of the action's cross-run state.
- **On by default for incremental native_loop** (per product decision), with `tool_evidence_memory: "false"` as a safety valve (the v1.2.10 lesson — never ship a behavior with no off-switch).
- **Staleness**: the digest is tagged with the SHA it was gathered at and framed as fail-safe context the model must re-verify against the delta — mirroring how carried findings must be re-confirmed.

## Safety

Injection posture mirrors `open_findings`: the digest draws from already-masked, size-capped tool output; it's written to a file (not eval'd) on parse, re-sanitized on load, and fenced as untrusted data in the corpus.

## Tests

- `test_evidence_memory.py` — build (final_text vs ledger, caps, control/angle stripping), load (sanitization, head_sha→hex, missing/blank), render (digest + gathered-at SHA + fail-safe framing).
- `test_carry_forward_roundtrip.sh` — extended with a full **marker → precheck → load** evidence path through the real code (persist, off-switch omits, empty omits, gathered-at SHA round-trips).
- `test_run_native_loop_wiring.py` — asserts the loop writes the digest into the result + `tool-harness.json`.
- **903 python + 25 smoke + shell suites green** (Python 3.14).

## Validation status

Pure-code + shell wiring is fully tested. Per the #265 testing bar, the **end-to-end reuse behavior is pending the #207 GPU eval** before merge (does seeding prior evidence reduce re-gathering without misleading the model on a stale fact?).

Closes the last #265 harness capability. Relates to #197, #265.


joryirving:
## Review response (pushed `ab24696`)

**Blocker — `check_review_needed.sh` regex is a false positive (verified), but tests added to lock it in.**

The flagged pattern `r'[\\x00-\\x08\\x0b-\\x1f<>]'` is **not** a no-op. The whole snippet runs inside a **bash double-quoted `python3 -c "..."`**, so bash collapses each `\\`→`\` *before* Python compiles the regex — Python receives `[\x00-\x08\x0b-\x1f<>]`, the correct control-char + angle-bracket class. It's also **byte-identical to the shipping `open_findings` sanitizer one line above** (already covered by the passing `angle brackets stripped from messages` round-trip assertion).

Proven empirically through the real extraction path — a hostile digest `fact v1.13.4 <BEL> <script>alert(1)</script> end` came out of `previous-evidence.json` as `fact v1.13.4  scriptalert(1)/script end` (control char gone, angle brackets gone, fact preserved). That case is now a **regression test** in `test_carry_forward_roundtrip.sh`, plus an inline comment so the doubling isn't re-flagged.

**Minor 1 — null-byte test:** added `test_strips_null_bytes_on_read` (`load_evidence_memory` re-strips `\x00` via `_CONTROL_CHARS_RE` — defense-in-depth even if the shell layer ever missed one).

**Minor 2 — symlink test:** added `test_symlinked_path_is_read_not_traversed`. Note this isn't a traversal vector: the path is hardcoded (`previous-evidence.json`) at every call site, as the review's own "path traversal: Not applicable" section confirms — the test just documents that a symlinked file still loads and is sanitized.

Net: the latent-risk concern (unsanitized JSON file) doesn't hold — the shell layer already strips control/markup chars, `load_evidence_memory` re-sanitizes, and the digest is fenced as untrusted in the corpus. No functional change; 905 tests + 25 smoke + shell suites green.

End-to-end reuse behavior remains gated on the #207 GPU eval before merge, as noted in the PR body.

PR #279, opened 2026-06-15
feat(tool-loop): keep tools on verdict turn for cache-prefix reuse (#263 Part 3)

### LINKED ISSUE #263: Native-loop prompt caching: cache-aware payload construction (per-format)

itsmiso-ai:
Fixes #263

## Summary

Part 3 of #263: keep `tools` present on the verdict turn so the cached prefix survives. Tool schemas render near the top of the prompt; dropping them invalidates the cache from that point onward.

## Changes

- **`conversation.py`**: Add `keep_tools_on_verdict: bool = False` parameter to `to_request_payload()`, `_to_openai_payload()`, and `_to_anthropic_payload()`. When True, tools are retained alongside `response_format` on the verdict turn.

- **`run_tool_harness.py`**: Enable `keep_tools_on_verdict=True` for the in-conversation verdict turn. The stable system prompt (Part 1) + tools (Part 3) means the full cached prefix survives across loop and verdict turns.

- **Tests**: 3 new unit tests (OpenAI, Anthropic, default behavior) + updated wiring test to expect tools on the verdict turn.

## Backend Compatibility

Most OpenAI-compatible backends support both `tools` + `response_format`. Some reject them together — when that happens, set `keep_tools_on_verdict=False` (the default). The primary deployment target (llama.cpp) supports this combination.

## Cache Impact

With Parts 1 + 2 + 3 all implemented:
- **System prompt**: stable across loop and verdict (no swap at token 0)
- **Tools**: present on every turn (prefix preserved)  
- **Anthropic**: `cache_control` markers on system + tools blocks

The verdict turn (the largest prefill) now reuses the cached prefix from loop turns instead of re-running from scratch.

### DIFF
--- a/pr_reviewer/conversation.py
+++ b/pr_reviewer/conversation.py
@@ -822,16 +822,18 @@ class Conversation:
         response_format: str | None = None,
         tokens_param: str = "max_tokens",
         cache_prefix: bool = False,
+        keep_tools_on_verdict: bool = False,
     ) -> dict[str, Any]:
         """Render the conversation as a wire-ready request body.
 
         Parameters mirror the bash ``build_model_request`` in
         ``scripts/model_call.sh`` so the loop driver can drop in with
         minimal reshuffling. ``verdict_turn=True`` triggers the
-        ``ai_response_format`` switch: ``tools`` is dropped, and the prior
-        conversation is either carried through (default off — see
-        ``keep_full_history_on_verdict``) or collapsed into a single system
-        note via :meth:`_verdict_transcript_note`.
+        ``ai_response_format`` switch: ``tools`` is dropped by default on the
+        verdict turn (set ``keep_tools_on_verdict=True`` to preserve them for
+        cache-prefix reuse — see #263 Part 3). The prior conversation is either
+        carried through (default off — see ``keep_full_history_on_verdict``) or
+        collapsed into a single system note via :meth:`_verdict_transcript_note`.
         """
         if api_format == "anthropic":
             return self._to_anthropic_payload(
@@ -843,6 +845,7 @@ class Conversation:
                 keep_full_history_on_verdict=keep_full_history_on_verdict,
                 response_format=response_format,
                 cache_prefix=cache_prefix,
+                keep_tools_on_verdict=keep_tools_on_verdict,
             )
         return self._to_openai_payload(
             model=model,
@@ -853,6 +856,7 @@ class Conversation:
             keep_full_history_on_verdict=keep_full_history_on_verdict,
             response_format=response_format,
             tokens_param=tokens_param,
+            keep_tools_on_verdict=keep_tools_on_verdict,
         )
 
     def _to_openai_payload(
@@ -866,6 +870,7 @@ class Conversation:
         keep_full_history_on_verdict: bool,
         response_format: str | None,
         tokens_param: str = "max_tokens",
+        keep_tools_on_verdict: bool = False,
     ) -> dict[str, Any]:
         system = self.system
         messages = self._render_openai_messages()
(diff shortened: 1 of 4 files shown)


joryirving:
Thanks for this — clean implementation, but closing: **Part 3 was empirically rejected by eval, not left unimplemented**, and #263 was closed as completed on that basis.

The reasoning here is half-right in a way that's worth flagging. You're correct that llama.cpp *accepts* `tools` + `response_format` together — but backend acceptance was never the blocker. The GPU eval (2026-06-14, qwen3.6-37b on llama.cpp, the deployment model class) found that **with tools present on the verdict turn the model emits a stray `tool_call` instead of the strict-JSON verdict — even when explicitly instructed to produce the verdict.** That breaks the verdict, so #205's design (drop tools on the verdict turn) is confirmed correct.

The unit tests here assert payload *shape* (`tools` present in the request) — they can't observe the model-behavior failure, which only the capability eval surfaced. And because `run_tool_harness.py` wires `keep_tools_on_verdict=True` on, this wouldn't be a dormant opt-in: it would regress the live verdict turn back into that failure mode.

The branch is preserved. If we ever move to a model that reliably emits JSON-with-tools-present, re-landing this would need a **capability eval (corpus-agentic pass rate), not unit tests**, to prove the verdict survives. Closing for now.

PR #348, opened 2026-06-24
fix(forgejo): send POST body via stdin; surface publish failures

eleboucher:
On the forgejo path, the publish step silently swallowed every API failure: the comment never posted, the job stayed green, no error in the log.

**Root cause**: `forgejo_backend.py:_curl` never passes `--data-binary @-` to curl on the dict-body path. POST/PATCH/PUT requests went out with an empty body, `create_comment` / `edit_last_comment` returned `None`, the CLI dispatcher printed `"null"` and exited 0, and `platform_api.sh` redirects stdout to `/dev/null`. Compounded by the dispatcher's lack of `sys.exit(1)` on `None` for those two subcommands — the other forgejo subcommands (`create-review-json`, `create-native-review`, `dismiss-review`) already do.

**Verified by**: a debug workflow step that ran a raw `curl` POST to the same endpoint with the same token — `http_status=201` immediately, confirming the API, URL, and token were all correct and the action's Python wrapper was the only thing broken.

**Three commits**:
1. `fix(forgejo): send POST body via stdin; surface publish failures` — `_curl` passes `--data-binary @-` unconditionally on the dict path; CLI `main()` `sys.exit(1)`s on `None` for `create-comment` and `edit-last-comment`
2. `feat(mcp): accept tool_name_prefixes to handle toolhive workload prefixes` — `is_read_only_tool` strips a known workload prefix (e.g. `github-mcp_`) before the leading-verb check, so aggregated MCP servers still register. Default-deny boundary preserved (tests assert `set_status`, `create_view`, `update_diff`, `refresh_status`, `delete_query`, `run_search` stay refused). Opt-in via new `tool_mcp_name_prefixes` input (default empty)
3. `docs(readme): document tool_mcp_name_prefixes input`

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,7 @@ Only three inputs are required: `github_token`, `ai_base_url`, and `ai_model`. E
 | `tool_enable_for_forks` | Allow tool harness on cross-repository PRs | No | `false` |
 | `tool_mcp_servers` | Allowlist of read-only MCP servers for `tool_mode: native_loop`, as a newline/comma list of `name=url`. Read-verb tools are advertised as `mcp__<name>__<tool>`; write-verb tools are refused. Empty = off. Fork-gated like the rest of the harness | No | `""` |
 | `tool_mcp_token` | Optional bearer token sent to every configured MCP server | No | `""` |
+| `tool_mcp_name_prefixes` | Comma-separated list of tool name prefixes to strip before the read-only verb check. Used to handle toolhive-style workload prefixes (`prefixFormat: "{workload}_"`) on aggregated MCP servers — e.g. `github-mcp,talos-mcp`. Empty preserves the default-deny boundary | No | `""` |
 
 </details>
 
--- a/action.yml
+++ b/action.yml
@@ -451,6 +451,10 @@ inputs:
     description: Optional bearer token sent to every configured MCP server (tool_mcp_servers)
     required: false
     default: ""
+  tool_mcp_name_prefixes:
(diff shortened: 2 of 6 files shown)


joryirving:
The `_curl` stdin fix and the `sys.exit(1)`-on-`None` additions look correct and match the existing pattern in the sibling subcommands. One concern on the third change.

## `is_read_only_tool` broadening weakens the default-deny boundary

The old check looked only at the leading verb (`split("_", 1)[0]`). The new logic accepts a tool if **any** of the first three `_`-segments is a read verb. Several `READ_ONLY_VERBS` (`status`, `view`, `diff`, `summary`, `query`, `head`, `stat`, `find`, `search`) commonly appear as the *second* word of a mutating tool name.

Concrete regression: an allowlisted MCP server advertising a write tool named `set_status`, `create_view`, `update_diff`, or `refresh_status`:

- **Before:** head is `set`/`create`/`update` → default-denied (not advertised, refused at call time).
- **After:** the second segment (`status`/`view`/`diff`) matches → the tool is advertised to the model *and* passes the call-time re-check, so the model can invoke a write operation.

That defeats the module's stated hard requirement (the docstring: "a tool is advertised only if its name begins with a read verb … anything else is default-denied"). The existing tests don't catch it — `set_y`, `run_z` have no read verb in any segment, so they still pass while the hole is open.

The change is also at the wrong altitude for its goal. The intent is to tolerate a known toolhive prefix (`github-mcp_…`), which is a fixed *leading* segment. The deeper fix is to strip the known prefix and then check the leading verb of the remainder — not scan-and-match-any across three positions, which admits read verbs that aren't the actual leading verb. A minimal safer form: check the first segment, and if it's a known prefix token, check the next one.

Suggest adding a test asserting `set_status` / `create_view` stay denied to lock the boundary.

PR #404, opened 2026-07-08
perf(planner): share section piece files with the corpus so verdict dedup drops them

### LINKED ISSUE #398: Align planner and corpus section builders so verdict-turn dedup can drop the big sections
Follow-up to #372 (partially fixed in #395).

#395 shipped byte-exact verdict-turn dedup, but measurement showed most planning-context sections are NOT byte-identical to their corpus counterparts, so the dedup can only drop small fully-duplicated sections (e.g. Version Hints):

- `PR Classification`: planner sends raw `classification.json` (4000B cap, json fence); corpus sends a jq projection, unfenced.
- `Changed Files` (planner) vs `PR Files (truncated)` (corpus): different titles and caps.
- `Repository Standards and Conventions` (planner, 6000B) vs `… ($STANDARDS_FILE)` (corpus, 16000B): different header and cap.
- `PR Diff (head)` (planner) vs `PR Diff (truncated)` (corpus): prefix overlap, which exact matching correctly refuses to drop.

**Fix:** make `build_planning_context` (`scripts/run_tool_harness.py`) and `build_review_corpus` (`scripts/sections/corpus.sh`) emit shared sections byte-identically (same title, projection, fence; caps reconciled or applied identically), so `dedupe_verdict_corpus` (already wired) drops them on the verdict turn — recovering most of the ~12k duplicated tokens #372 measured.

**Caution:** this changes the model's planning input, not just the verdict turn — needs N≥3 cache-busted eval against the home-ops#7462 fixture before shipping. Post-release work, not a hotfix.

joryirving:
Closes #398. **Draft until the N=3 live eval against home-ops#7462 passes** (model-input change).

## Summary
- `build_review_corpus` writes four shared piece files (classification, PR files, version hints, standards) containing the exact bytes it already emits for those sections, then assembles the corpus from them. **Corpus output is byte-identical** — verified empirically by extracting old (main) and new `build_review_corpus`, running both over identical fixtures (multibyte content, both scopes), and `cmp`-ing: full and incremental match byte-for-byte.
- `build_planning_context` embeds a shared piece verbatim when the whole piece fits its budget (2200-byte reserve keeps the guaranteed diff head), else falls back to exactly today's excerpts (different titles/caps, so dedup conservatively keeps the corpus copy). The diff head is never shared — prefix overlap isn't byte-dedupable.
- Net effect in native_loop: the #395 dedup now drops classification + version hints + standards from the uncached verdict turn; their bytes move into message 1, which the loop's prompt cache covers across rounds.

## Verification
- 1210 pytest + all `tests/test_*.sh` green.
- New: `test_corpus_shared_pieces.sh` (pins piece bytes exactly, asserts verbatim presence in the corpus, exercises dedup on the real pieces); 4 planner tests (fit→embed+dedup, oversize→excerpt fallback, missing piece→fallback, diff-head cap).
- Pending: N=3/arm live eval on home-ops#7462 (verdicts, findings, tool traces, prompt tokens) — results will be posted here before undrafting.

### DIFF
--- a/scripts/run_tool_harness.py
+++ b/scripts/run_tool_harness.py
@@ -156,43 +156,76 @@ def build_planning_context(max_bytes, corpus_path=None):
     and the head of the diff. Falls back to the corpus head when the piece
     files are unavailable (e.g. standalone invocation).
 
+    Shared-piece preference (#398): each high-signal section has a shared piece
+    file written ONCE by build_review_corpus (scripts/sections/corpus.sh) that
+    holds the EXACT bytes the corpus emits for that section. When a shared piece
+    exists and fits WHOLE within the remaining budget, we embed it verbatim
+    (rstripped) — title, fences and all — so the verdict-turn dedup
+    (dedupe_verdict_corpus) can drop the byte-identical corpus copy. Full-fit
+    only: a partial/clipped embed would never match the corpus section verbatim
+    and would defeat the byte-exact dedup, so we fall back to the older
+    self-built excerpt (different title/cap/fence) instead — the excerpt then
+    just fails to dedup and the corpus section is (correctly, conservatively)
+    kept in full. A fixed 2200-byte reserve (the guaranteed 2000-byte diff head
+    + margin) is withheld from every shared-fit check so a large piece can't
+    starve the diff.
+
     Returns (text, truncated).
     """
+    # (planner_title, shared_piece_path, excerpt_source_path, excerpt_cap, fence)
     pieces = [
-        ("PR Classification", "classification.json", 4000, "json"),
-        ("Changed Files", "pr-files.truncated.json", 6000, "json"),
-        ("Version Hints from Diff", "version-hints.truncated.txt", 2500, "text"),
-        ("Repository Standards and Conventions", "standards-context.capped.md", 6000, None),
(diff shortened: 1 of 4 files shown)


joryirving:
Eval gate resolved with offline evidence instead of live model runs — the change turned out lower-risk than the issue anticipated.

Real fixture data (home-ops#7462, production config: 15KB planning budget, AGENTS.md standards) through both versions:

**Message-1 delta (the only model-visible change):**
- Classification section: pretty-printed raw JSON → the compact corpus projection (same fields/values; drops only `route_signals`, which is routing metadata).
- `# Changed Files` → `# PR Files (truncated)` (title only, same content).
- Standards: the full 8.7KB piece fits and embeds whole, replacing the 6KB excerpt — the planner sees *more* standards, not less.
- Net planning-context size: 14,733 → 14,841 bytes (+108B). Diff head present in both.

**Verdict-turn effect:** 20,830 → 9,582 bytes (6 sections dropped, vs 2 with the old planning context) — 54% smaller uncached tail on this small fixture; scales with PR size.

Given the delta is a formatting change plus strictly-more standards context, skipping the N=3 live runs. Corpus byte-identity is separately proven (old-vs-new `cmp`, both scopes).

PR #410, opened 2026-07-11
fix: omit Linked Issue Context / Evidence Providers headers when empty

Tanguille:
Closes #409, completes #399.

## What

`corpus.sh` unconditionally printed both the `# Linked Issue Context` and
`# Evidence Providers` headers regardless of whether the underlying file
had any content. #400 fixed `context.sh` to leave `linked-issues.md`
empty when there's no linked issue, but the header above it in
`corpus.sh` stayed unconditional — so the model still saw a section
boundary with nothing under it, and (per #399's original report) reliably
filled it in anyway rather than omitting the section.

`run_evidence_providers.py` had the identical bug in a second code path
#400 never touched: the "not configured" branch wrote a literal
`"No evidence providers configured."` sentence into `evidence-providers.md`,
which `corpus.sh` then unconditionally headered.

## Fix

- `corpus.sh`: gate both headers on `[ -s <file> ]`, the same pattern
  already used for the CI Check Results section a few lines below.
- `run_evidence_providers.py`: only the "not configured" branch (no
  `EVIDENCE_PROVIDERS_FILE` set at all) now writes an empty markdown file.
  The "config not found" and "invalid JSON" branches are left untouched —
  those are real misconfigurations worth surfacing, not "nothing to
  report", so they should stay visible under the (now correctly gated)
  header.

## Testing

- Added `test_not_configured_writes_empty_markdown` and
  `test_config_not_found_still_reports_error` to
  `tests/test_evidence_providers.py`, both passing.
- Full suite: `pytest tests/` — 1212 passed. The 2 failures in
  `TestMarkdownCaps` are pre-existing on unmodified `main` in my sandbox
  (a local `python3 -c` subprocess-interception quirk unrelated to this
  change — verified by stashing and re-running against the base commit).
- `tests/test_advisory_parallel.sh` — 18 passed, 0 failed.
- `bash -n scripts/sections/corpus.sh` — clean.

## Evidence

Real review output before this fix, on a routine dependency-bump PR with
no linked issue and no evidence providers configured:

> ### Linked Issue Fit
> No linked issues were provided for this PR. As a routine dependency update, no specific issue acceptance criteria need to be verified.
>
> ### Evidence Provider Findings
> No evidence provider findings were generated.

Both sections should be entirely absent per `default_system_prompt.txt`'s
own stated conditional ("a Linked Issue Fit section **when issue context
is available**", "an Evidence Provider Findings section **when provider
output exists**").

PR #416, opened 2026-07-12
fix: omit empty Linked Issue Fit / Evidence Provider Findings sections (#415)

### LINKED ISSUE #415: Review output still emits filler for empty Linked Issue Fit / Evidence Provider Findings sections despite #410
## Follow-up to #409 / #410

#410 fixed the *input* side: `corpus.sh` no longer hands the model an empty "# Linked Issue Context" / "# Evidence Providers" header when the underlying file is empty.

That's insufficient on its own. Observed on a live v2.1.3 review (PR with no linked issue, no evidence providers configured):

```
## Linked Issue Fit

No linked issue was present. The PR is a standard Renovate-driven dependency update.

## Evidence Provider Findings

No evidence provider findings were present.
```

Root cause: `default_system_prompt.txt` tells the model to include these output sections "when available" — a soft conditional it doesn't reliably obey, even with a clean (header-free) corpus. The model confabulates the section and a "nothing was found" filler sentence anyway.

Fix incoming in a PR: strengthen the instruction, and — since prompt-only fixes for this exact failure mode already didn't hold once — add a deterministic post-processing step (mirroring `strip_metadata_markers.py`) that mechanically strips these sections from `review_markdown` using the same presence signals `corpus.sh` already computes.

itsmiso-ai:
Issue #415: even after #410 cleaned up the input side (no empty headers in the corpus), the model still confabulated `## Linked Issue Fit` / `## Evidence Provider Findings` sections padded with 'nothing was found' filler. The root cause per the issue is twofold: a soft prompt conditional the model doesn't reliably obey, *and* the absence of a deterministic post-processing step to enforce it mechanically.

This PR implements both fixes the issue requires:

**1. Strengthen the instruction** (`default_system_prompt.txt`) — the per-section omit directive now names each conditional section's own trigger (Linked Issue Fit → issue context, Evidence Provider Findings → provider output, Tool Harness Findings → tool output) and explicitly forbids `"- findings: []"` filler. Prompt-only fixes for this exact failure mode recurred once already, so this alone is insufficient.

**2. Deterministic publish-time backstop** — new `scripts/strip_empty_conditional_sections.py`, mirroring `strip_metadata_markers.py`. Removes a `## Linked Issue Fit` / `## Evidence Provider Findings` heading and its body (through the next same/higher-level heading or EOF) when that section's corpus signal is absent. Wired into `sanitize_review_markdown` so it runs in all three publish modes. Presence is derived from the *exact* ``[ -s linked-issues.md ]`` / ``[ -s evidence-providers.md ]`` gates `corpus.sh`` uses to emit (or omit) those section headers — so a section is only ever stripped when the corpus never offered it. Fail-safe (missing signal → keep), fenced-code-block aware (won't match `#` inside code).

**Tests:** `tests/test_strip_empty_conditional_sections.py` (13 cases, TDD red-then-green) covers the exact #415 bug, both sections, selectivity (one present / one absent), sub-heading removal, code-fence boundary, and layout hygiene. Full suite 1227 passed; `test_system_prompt_fragments.sh` 29 passed.

Fixes #415


itsmiso-ai:

Posted automatically by Dispatch on 2026-07-13T02:42:52.218Z

joryirving:
Response to the review (request_changes):

**Finding 1 (blocker / security) — disputed.** Verified each sub-claim against the code before acting:

- **Null bytes:** the claim that `## Linked Issue Fit\x00Hack` "could cause the section not to be stripped" is incorrect. `.+?` matches the null byte, `str.lower()` preserves it, and `startswith("linked issue")` returns `True` — the section is stripped. Verified:
  ```
  matched=True group2="Linked Issue Fit\x00Hack" startswith=True
  stripped contains "Linked Issue Fit": False   # WAS stripped
  ```
- **Symlinks:** already defended upstream. `assert_safe_artifact_paths` (scripts/artifact_paths.sh:20-21) refuses to write through symlinks for the exact artifact paths this script operates on (`review-markdown.raw.md`, `review-comment-markdown.raw.md`, `review-verdict-markdown.raw.md`). The script's only caller is `sanitize_review_markdown`, which passes a local var, not untrusted input. The model cannot create symlinks in the CI checkout.
- **Bidi override suffix:** `## Linked Issue Fit\u202e` → `startswith=True` → stripped fine (verified).
- **7 hashes:** `####### H7` is not a heading per CommonMark (max 6), so treating it as body text is correct, not a leak.
- **Zero-width-space prefix (`\u200bLinked Issue Fit`):** this one is real — it slips through `startswith`. But it is cosmetic (filler shows in the comment), not a security issue, and requires the model to emit a leading zero-width space, which is not an observed failure mode. Not addressing it as part of this fix.

The `path_handling_changes` flag and its must_check are a generic classifier heuristic that fires on any script accepting a file-path argument; this script is not a path-handling security boundary.

**Finding 2 (minor / docs) — accepted.** Fixed: AGENTS.md entry now documents the fence-aware safety property.

**Must_check closure:** added `TestEdgeCaseInputs` (f99946f) pinning the null-byte and 7-hash behavior with regression tests so the must_check is answered with evidence rather than assertion. Suite: 1230 passed.

PR #446, opened 2026-07-30
fix(review): prevent stale and unpublishable AI verdicts

phorrr:
## Summary

- fail before model invocation when Forgejo PR read/publication access is unavailable
- skip superseded PR heads before model use and recheck immediately before publishing
- bind GitHub and Forgejo native reviews to the reviewed commit SHA
- publish a non-blocking advisory `COMMENT` when a clean verdict cannot safely approve
- fall back to an advisory comment if native approval fails
- redact Forgejo publication diagnostics
- standardize the self-review workflow on the released v2.1.7 baseline and low-retry/fork-safe settings

## Why

A private Forgejo consumer spent on model reviews and only then failed publication because its reviewer token lacked repository access. Separately, a push during model execution could allow a verdict generated for an old head to be published against the newer PR, and clean fork reviews withheld from approval were represented as false `REQUEST_CHANGES` verdicts.

## Validation

- `uvx --from pytest pytest tests/ -q`: 1276 passed
- all 29 `tests/test_*.sh` suites passed
- `git diff --check`
- Python compile and shell syntax checks
- consumer/config-service compatibility was validated against the composite action metadata in a separate integration pass

## Rollout

After merge, release the next patch and update immutable consumer pins. Existing v2.1.7 consumers remain compatible.

### DIFF
--- a/.github/workflows/ai-pr-review.yaml
+++ b/.github/workflows/ai-pr-review.yaml
@@ -42,22 +42,23 @@ jobs:
         uses: misospace/pr-reviewer-action@258ac494ff48261cba9a62dce384b35ed73404ba # v2.1.7
         with:
           github_token: ${{ steps.app-token.outputs.token }}
-          ai_primary_retries: "3"
-          ai_primary_retry_delay_sec: "15"
+          ai_primary_retries: "2"
+          ai_primary_retry_delay_sec: "5"
+          ai_connect_timeout_sec: "10"
           ai_base_url: ${{ vars.LITELLM_URL }}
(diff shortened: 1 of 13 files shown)


phorrr:
CI validation is green (1276 Python tests plus all shell suites). The AI Review job fails before checkout because GitHub correctly withholds `BOT_CLIENT_ID`/app secrets from external fork PRs; it never invokes the reviewer. A maintainer-side review/merge is required before the next patch can be released and downstream immutable pins updated. @joryirving

PR #455, opened 2026-07-30
fix(forgejo): accept owner publication permission

phorrr:
## Summary

Normalize Forgejo's `owner` repository permission to `admin` during the publication-access preflight.

## Problem

Forgejo may return `"permission": "owner"` for organization owners. The preflight currently recognizes only `read`, `write`, and `admin`, so a fully authorized organization owner is treated as having unknown permissions and the review exits before model invocation.

## Changes

- Map Forgejo's `owner` permission to the action's existing `admin` permission level.
- Add regression coverage for the organization-owner response.
- Preserve fail-closed behavior for all other unknown permission values.

## Validation

- 1,281 tests passed across the full Python test suite.
- Verified against a live Forgejo organization-owner token: preflight and review publication completed successfully.

### DIFF
--- a/pr_reviewer/forgejo_backend.py
+++ b/pr_reviewer/forgejo_backend.py
@@ -248,6 +248,10 @@ def get_authenticated_repo_permission(repo_full_name: str) -> str | None:
     )
     data = _json_decode(body_text)
     permission = data.get("permission") if status_code == 200 and isinstance(data, dict) else None
+    # Forgejo reports organization owners as "owner" rather than "admin".
+    # Normalize both to the action's coarse read/write/admin vocabulary.
+    if permission == "owner":
+        return "admin"
     if permission not in {"read", "write", "admin"}:
         _report_http_error("review access preflight", status_code, body_text)
         return None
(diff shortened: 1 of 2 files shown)


pr_reviewer/forgejo_backend.py
```diff
@@ -248,6 +248,10 @@ def get_authenticated_repo_permission(repo_full_name: str) -> str | None:
     )
     data = _json_decode(body_text)
     permission = data.get("permission") if status_code == 200 and isinstance(data, dict) else None
+    # Forgejo reports organization owners as "owner" rather than "admin".
+    # Normalize both to the action's coarse read/write/admin vocabulary.
+    if permission == "owner":
+        return "admin"
```
joryirving: Non-blocking: reassigning instead of early-returning keeps a single exit path, so the accepted-set check below stays the only gate (and personal repo owners report `owner` too, not just org owners):

```suggestion
    # Forgejo reports repository/organization owners as "owner" rather than
    # "admin". Normalize to the action's coarse read/write/admin vocabulary.
    if permission == "owner":
        permission = "admin"
```

joryirving (dismissed):
Confirmed against the Gitea/Forgejo `AccessMode` lineage: `owner` sits above `admin` and is exactly what self-hosted owner tokens report, so #454's preflight would have falsely rejected the most-privileged case. Fail-closed behavior for other unknown values is preserved and the regression test covers the reported failure. One non-blocking style suggestion inline.

Issue #231, opened 2026-06-12
Manual "force re-review" override (Renovate-style checkbox / label) to bypass the diff-unchanged guard


joryirving:
## Problem

The precheck skips a review when the **broad fingerprint** (diff `patch-id` + config hash) matches the one embedded in the last managed comment:

- Fingerprint computed in [`scripts/check_review_needed.sh`](scripts/check_review_needed.sh) — diff `git patch-id --stable` (~L19-30) combined with `compute_config_hash()` (~L32-192) into `"{patch_id}|cfg:{config_hash}"` (~L195).
- Skip decision (~L236-242): when `skip_if_diff_unchanged=true` and `last_broad_fingerprint == broad_fingerprint`, sets `should_review=false` (`skip_reason="diff-unchanged"`) and short-circuits.

This is correct almost always. But it makes a **manual re-run of the workflow a no-op** when neither the diff nor the config string changed — and there are real cases where you *want* a fresh review of an unchanged PR:

- **The model changed behind a stable alias.** Our fleet points `ai_model` at a LiteLLM alias (e.g. `review`). When the alias is repointed to a different backend model, the config-hash inputs are byte-identical, the fingerprint is stable, and a re-run refuses to re-review — even though the reviewer would now behave differently.
- Transient model flakiness / a bad first review you want to regenerate.
- Standards or prompt changes that live behind indirection the config hash doesn't capture.

Today the **only** ways to force a review are to commit a dummy change (busts the diff patch-id) or flip `skip_if_diff_unchanged: false` in the workflow (a config edit that re-reviews *everything*, fleet-wide). Both are clumsy.

## Existing escape hatches (for reference)

- `skip_if_diff_unchanged` input ([action.yml](action.yml) ~L383-386) — workflow-level, not per-PR, all-or-nothing.
- `workflow_dispatch` trigger exists but takes **no inputs** and the action step is gated `if: github.event_name == 'pull_request'` ([ai-pr-review.yaml](.github/workflows/ai-pr-review.yaml) ~L7, ~L36), so a manual dispatch still respects the guard.
- No label check, no magic-comment string, no per-PR override.

## Proposal

Add a per-PR **"force re-review"** override, modeled on Renovate's dependency-dashboard checkbox. The flow the user described: *check the box, re-run the workflow, and it actually re-reviews even though nothing changed* — then the box resets so it doesn't loop.

### Mechanism (one-shot, persisted in the managed comment)
[...]

PR #502, opened 2026-08-14
Finish precheck migration: have check_review_needed.sh delegate to precheck.py

### LINKED ISSUE #497: [P3] precheck.py ships ~300 lines of tested decision logic the action never executes — dual scope/fingerprint implementations with the shell
Ask: Finish the precheck migration by having check_review_needed.sh execute precheck.py's decision functions and delete the shell duplicate.
Expected files: pr_reviewer/precheck.py, scripts/check_review_needed.sh

**Problem:** The #429/#482 migration moved `check_review_needed.sh` logic into `pr_reviewer/precheck.py` — but only `compute_config_hash()` and `fingerprints_match()` are actually called by the shell wrapper. The other ~300 lines of exported, unit-tested decision logic (`should_review`, `_detect_incremental_scope`, `_parse_diff_stats`, `compute_diff_fingerprint`, `extract_config_lines`, `_format_output`, `PrecheckResult`, `main`) are exercised **only by tests**. Meanwhile the live action path keeps its own parallel implementations in bash: incremental-scope resolution (`resolve_review_scope`, `extract_review_metadata`), diff fingerprinting via `git patch-id --stable`, and the should-review decision. The action now has two implementations of the same review-scope/fingerprint logic that can silently diverge — the exact drift class #482 was meant to eliminate — and fixing the Python copy gives a false sense of progress because nothing in production runs it.

**Evidence:**
- `scripts/check_review_needed.sh:82-83` and `:131-135` call only `compute_config_hash` and `fingerprints_match` from `pr_reviewer.precheck`.
- `grep -rn "should_review\|_detect_incremental_scope\|compute_diff_fingerprint\|_parse_diff_stats\|_format_output\|PrecheckResult\|extract_config_lines" scripts/ action.yml` → no call sites outside `pr_reviewer/precheck.py` and `tests/unit/test_precheck.py` (the test file imports and asserts all of them at lines 20-32).
- Shell-side live logic that duplicates the Python: `scripts/check_review_needed.sh:199-298` (`resolve_review_scope`/`extract_review_metadata`) and `:66-67` (`git patch-id --stable < pr.diff`), while `precheck.py:67-123` has `compute_diff_fingerprint` (SHA-256, different algorithm) and `:345-459` has `_detect_incremental_scope`/`should_review` with the same `MAX_INCREMENTAL_FILES/MAX_INCREMENTAL_LINES/MIN_INCREMENTAL_RATIO` constants mirrored from the shell.
- `precheck.py:576` `main()` has no caller anywhere in scripts/ or action.yml.

**Acceptance:**
- `check_review_needed.sh` derives should-review, scope, and fingerprint decisions from `precheck.py` (or the unused Python functions are deleted and the shell stays the single source).
- Exactly one implementation of scope detection and fingerprinting remains in the action path, and `tests/unit/test_precheck.py` exercises the surviving implementation.
- The diff-unchanged skip and incremental-scope behaviour in `test_review_routing.sh` / `test_review_scope.sh` is unchanged.

---

## Current state — read this before starting

PR #502 is already open for this issue on branch `foreman/wl-misospace-pr-reviewer-action-497/issue-497`, head `dd1a444`. **Continue that branch. Do not start over.** Two previous attempts rewrote `scripts/check_review_needed.sh` wholesale and each time lost working code. The branch is fully in sync with `main` (0 commits behind).

### Already fixed — do NOT undo any of this

An earlier round deleted load-bearing code along with the logic being migrated. It has since been restored, and `tests/test_artifact_paths.sh` passes as a result. Keep all of it:

- `OUTPUT_FILE="${GITHUB_OUTPUT:-...}"`. `action.yml:672` names the step `id: precheck` and 30 references read 11 outputs from it (`should_review`, `skip_reason`, `effective_review_scope`, `previous_head_sha`, `baseline_clean`, `head_sha`, `base_sha`, `is_fork_pr`, `diff_fingerprint`, `resolved_platform`, `effective_forgejo_api_url`). Writing anywhere other than `$GITHUB_OUTPUT` makes every one of them empty and the action stops reviewing anything.
- `source "${SCRIPT_DIR}/platform_api.sh"` and `source "${SCRIPT_DIR}/artifact_paths.sh"`. There is no `scripts/_lib.sh` in this repo and never has been.
- The env-var preamble setting `REPO`, `PR_NUMBER`, `COMMENT_MARKER`, `PUBLISH_MODE`, and the `REPO`/`PR_NUMBER` guard. The script runs under `set -euo pipefail`, so a missing default is an unbound-variable exit before any logic runs.
- `assert_safe_artifact_paths`, `REREVIEW_LABEL` (the #231 label-driven re-review), `platform_resolve`, `SKIP_IF_DIFF_UNCHANGED`.
- The `PYTHONPATH` export matching `run_review.sh:13`.

### What is still broken

Three suites fail on `dd1a444`. These are decision-logic bugs in the migration itself, not plumbing.

`tests/test_check_review_needed.sh` — 39 passed, 8 failed. Every failure is the skip path never triggering:

```
FAIL: should_review=false when broad fingerprint matches (got 'true', expected 'false')
FAIL: should_review=false without force (matching fingerprint)  (got 'true', expected 'false')
FAIL: should_review=false when everything matches              (got 'true', expected 'false')
FAIL: should_review=false when a PR review fingerprint matches (review_verdict)
FAIL: should_review=false on matching fingerprint              (got 'true', expected 'false')
FAIL: PR object is not fetched when skipping                   (got '1', expected '0')
FAIL: broad fingerprint has pipe delimiter                     (got 'no', expected 'yes')
FAIL: empty diff fingerprint contains 'empty-diff'             (got 'no', expected 'yes')
```

**Start with the last two — they are almost certainly the cause of the other six.** `build_broad_fingerprint` must produce `<diff_fp>|cfg:<config_hash>`, and an empty diff must produce a fingerprint containing `empty-diff`. Neither marker appears in what the shell currently exports, so the value being compared against the stored `<!-- ai-pr-review-fingerprint:... -->` marker is not in that marker's format and no comparison can ever match. A fingerprint that never matches means the action re-reviews every push forever. `PR object is not fetched when skipping` is the same bug from the other side: the skip short-circuit is never reached, so the fetch it should have avoided still happens.

`tests/test_review_scope.sh` — 18 passed, 3 failed:

```
FAIL: non-forced auto review uses incremental scope with a safe baseline (got 'full', expected 'incremental')
FAIL: safe incremental review trusts clean baseline                      (got 'false', expected 'true')
FAIL: non-forced dirty-baseline review remains incremental               (got 'full', expected 'incremental')
```

`_detect_incremental_scope` is either not reached or its result never reaches `effective_review_scope` and `baseline_clean`. The shell passes `ENABLE_INCREMENTAL_DETECTION`, `PREVIOUS_HEAD_SHA` and `PREVIOUS_BASE_SHA` into the Python — confirm all three arrive with the values the tests set up, and that the resulting scope is written to `$GITHUB_OUTPUT` and not only into `precheck-result.json`.

`tests/test_carry_forward_roundtrip.sh` — fails because `extract_review_metadata` was deleted. That suite extracts the function out of `check_review_needed.sh` by regex at line 35 and sources it, so it asserts against real code that no longer exists. **Pick one and say which in the PR description:** either keep `extract_review_metadata` as a shell function so the roundtrip test keeps exercising the live implementation, or move its logic into `precheck.py` and rewrite that test to call the Python directly. Both are acceptable. Adding a dead bash copy that exists only to satisfy the regex is not.

### Also outstanding

- `shellcheck` SC2155 at `scripts/check_review_needed.sh:199`, `:203`, `:204` — declare and assign separately.
- A duplicate assignment of `previous_fingerprints_raw` before the `declare -F` guard is always overwritten by the guarded one below it. Remove the first.

### Definition of done

`tests/test_check_review_needed.sh`, `tests/test_review_scope.sh` and `tests/test_carry_forward_roundtrip.sh` all green, shellcheck clean, and the 11 outputs above still written to `$GITHUB_OUTPUT`. Do not weaken any test assertion to achieve this — the assertions are measuring whether the migrated implementation makes the same decisions the old one did, which is the whole point.

itsmiso-ai:
## What
The script check_review_needed.sh now delegates diff fingerprinting, config hash, and review-scope decisions to pr_reviewer.precheck via `python3 -m pr_reviewer.precheck`, sources the structured precheck.out to obtain DIFF_FINGERPRINT, CONFIG_HASH, BROAD_FINGERPRINT, D…

Fixes #497

_Opened by foreman on review GO (workload wl-misospace-pr-reviewer-action-497)._

### DIFF
--- a/pr_reviewer/precheck.py
+++ b/pr_reviewer/precheck.py
@@ -4,6 +4,42 @@ Core functions for diff fingerprinting, incremental scope detection,
 config hash computation, and review metadata transport.
 
 These replace the shell implementations with testable Python code.
+
+CLI contract (``python3 -m pr_reviewer.precheck``)
+---------------------------------------------------
+
+The module entry point (``main``) reads its inputs from the environment
+and writes a single JSON object to stdout with the keys ``should_review``,
+``skip_reason``, ``effective_review_scope``, ``previous_head_sha``,
+``baseline_clean``, ``diff_fingerprint``, ``broad_fingerprint`` and
+``config_hash``. It is the decision half of the action precheck; the shell
+wrapper performs platform I/O (diff/PR/comment fetches) and forwards the
+result to ``$GITHUB_OUTPUT``.
+
+Environment inputs:
+
+- ``PRECHECK_DIFF_PATH`` / ``DIFF_PATH``: path to the PR diff file
+  (``PRECHECK_DIFF_PATH`` wins; ``DIFF_CONTENT`` is the fallback when
+  neither file is set).
+- ``PREV_FINGERPRINTS``: comma-separated fingerprints stored in previous
+  managed reviews (``PREV_FP_PATH`` file with ``diff-fp:`` lines is also
+  honoured).
+- ``FORCE_REVIEW``: ``true`` bypasses the diff-unchanged guard.
+- ``SKIP_IF_DIFF_UNCHANGED``: ``true`` (default) enables the guard.
+- ``REVIEW_SCOPE``: user scope request (``full`` / ``incremental`` /
+  ``auto``); ``auto`` is the default.
+- ``PREVIOUS_HEAD_SHA`` / ``PREVIOUS_BASE_SHA`` / ``PREVIOUS_REVIEW_RESULT``:
+  metadata from the last managed review.
+- ``PREVIOUS_HEAD_IS_ANCESTOR`` / ``COMPARE_RANGE_OK``: caller-supplied
+  validation verdicts for the previous→current range (ancestorship and
+  compare/base continuity). Unset means "not asserted" (the check does
+  not gate the baseline); ``false`` forces a full-scope fallback.
+
+``diff_fingerprint`` is the diff's own fingerprint (``empty-diff``
+placeholder for an empty diff); ``broad_fingerprint`` is the marker form
+``<diff_fingerprint>|cfg:<config_hash>`` stored in the
+``ai-pr-review-fingerprint`` comment marker, so a published fingerprint
+round-trips into the diff-unchanged comparison on the next run.
 """
 
 import hashlib
@@ -27,6 +63,14 @@ FP_PREFIX = "diff-fp:"
 
 # Delimiter separating diff fingerprint from config hash in broad fingerprint
 FP_DELIMITER = "|"
+
+# Placeholder fingerprint for an empty diff, so the stored-marker
+# comparison has a stable value to match on re-runs.
+EMPTY_DIFF_FINGERPRINT = "empty-diff"
+
+# Suffix marking the config-hash half of a marker fingerprint, as stored in
+# the ai-pr-review-fingerprint comment marker: `<diff_fp>|cfg:<config_hash>`.
+CONFIG_HASH_MARKER = "cfg:"
 
 # Incremental scope detection constants
 MAX_INCREMENTAL_FILES = 20
@@ -57,6 +101,19 @@ class PrecheckResult:
     total_files: int = 0
     total_lines: int = 0
     reason: str = ""
+
+
+@dataclass
+class ScopeResolution:
+    """Outcome of metadata-based review scope resolution.
+
+    A full scope always clears the carried incremental baseline: no
+    previous head is forwarded and the baseline is untrusted.
+    """
+
+    effective_review_scope: str = "full"
+    previous_head_sha: str = ""
+    baseline_clean: bool = False
 
 
 # ---------------------------------------------------------------------------
@@ -242,6 +299,36 @@ def build_broad_fingerprint(diff_fp: str, config_hash: str) -> str:
     if not config_hash:
         return diff_fp
     return f"{diff_fp}{FP_DELIMITER}{config_hash}"
+
+
+def build_marker_fingerprint(diff_fp: str, config_hash: str) -> str:
+    """Build the marker-compatible broad fingerprint.
+
+    Unlike :func:`build_broad_fingerprint` (the library-internal
+    ``<diff_fp>|<config_hash>`` form), this is the exact string stored in
+    the ``ai-pr-review-fingerprint`` comment marker and compared against
+    ``PREV_FINGERPRINTS`` on the next run:
+
+    - an empty ``diff_fp`` becomes the ``empty-diff`` placeholder, so an
+      empty diff still yields a stable, matchable fingerprint;
+    - the config hash is always present in ``|cfg:<hash>`` form, matching
+      the shell's synthesis and keeping the marker parseable even when the
+      hash is empty.
+
+    Parameters
+    ----------
+    diff_fp : str
+        Diff fingerprint (empty string for an empty diff).
+    config_hash : str
+        Config hash from ``compute_config_hash`` (may be empty).
+
+    Returns
+    -------
+    str
+        ``<diff_fp or empty-diff>|cfg:<config_hash>``.
+    """
+    fp = diff_fp or EMPTY_DIFF_FINGERPRINT
+    return f"{fp}{FP_DELIMITER}{CONFIG_HASH_MARKER}{config_hash}"
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


itsmiso-ai:

**Reason:** Failing check: validate (failure)

**Latest note:** foreman fix Workload prfix-misospace-pr-reviewer-action-502 succeeded but PR is still not mergeable after 3/3 attempts

Posted automatically by Dispatch on 2026-08-14T09:30:10.033Z

joryirving (changes requested):
Request changes. The Python delegation to `pr_reviewer.precheck` is the right core and should be kept, but the rewrite of `scripts/check_review_needed.sh` cut the file from 358 lines to 126 and took out load-bearing code along with the logic it was meant to replace. Three test suites fail on this branch and CI on `main` is green, so these are regressions introduced here.

The most serious problem is not covered by any test. `action.yml:672` gives the step `id: precheck`, and 30 references across `action.yml` read 11 distinct outputs from it: `should_review`, `skip_reason`, `effective_review_scope`, `previous_head_sha`, `baseline_clean`, `head_sha`, `base_sha`, `is_fork_pr`, `diff_fingerprint`, `resolved_platform`, `effective_forgejo_api_url`. Step outputs only exist if they are written to `$GITHUB_OUTPUT`. On `main` that is why line 16 reads `OUTPUT_FILE="${GITHUB_OUTPUT:-/dev/null}"`. This branch changed it to `OUTPUT_FILE="${OUTPUT_FILE:-check_review_needed.out}"`, a plain file in the working directory that the runner never reads, so every one of those 11 outputs resolves to the empty string and the action stops reviewing anything. Restore the `$GITHUB_OUTPUT` target.

The script also cannot start. It reads `REPO` (4 times), `PR_NUMBER` (4), `COMMENT_MARKER` (3) and `PUBLISH_MODE` (1), but the preamble that set all four was deleted, and the file runs under `set -euo pipefail`, so the first expansion exits with an unbound variable error before any logic runs. `main` sets them at lines 10-18 with the usual `${VAR:-default}` pattern. Restore that block, including the `REPO`/`PR_NUMBER` guard at lines 26-29.

Line 14 sources `"$SCRIPT_DIR/_lib.sh"`. That file does not exist on this branch, on `main`, or anywhere in the repo's history; the only `_lib` is `tests/_lib/assert.sh`, which is a test helper. `main` sourced two real files here instead, `platform_api.sh` at line 22 and `artifact_paths.sh` at line 24, and the script needs the first one for `platform_issue_comments`, `platform_pr_reviews` and `platform_pr_diff`, all three of which it still calls. This is what makes `tests/test_check_review_needed.sh` fail with `_lib.sh: No such file or directory`. Replace the bogus source with those two.

Beyond that, several features were removed with no replacement in `precheck.py`. `REREVIEW_LABEL` had 4 references and implemented the label-driven re-review from #231, including the "not the re-review trigger" notice; all of it is gone. `platform_resolve` is gone, yet the script still emits `resolved_platform=${RESOLVED_PLATFORM:-}`, so that output is now permanently empty and the Forgejo branch of `action.yml` can never be taken. `SKIP_IF_DIFF_UNCHANGED` is a documented input that is now read nowhere and silently ignored. `assert_safe_artifact_paths` is gone, which `main` calls at line 32 under the comment "Guard artifact paths before any redirect can follow a PR-controlled symlink"; its removal is what fails `tests/test_artifact_paths.sh:59`. `resolve_review_scope` and `fallback_full_scope` accounted for 9 references between them and are gone as well.

`extract_review_metadata` needs a decision rather than a straight restore. `tests/test_carry_forward_roundtrip.sh:35` extracts that function out of `check_review_needed.sh` by name with a regex and sources it, deliberately exercising the real implementation. The function is gone, so the suite fails. Either keep it in the shell so the roundtrip test keeps testing real code, or move its logic into `precheck.py` and rewrite that test to exercise the Python. Do not add a dead bash copy that exists only to satisfy the regex; if you take the Python route, say so in the PR description so the test change is reviewable on purpose.

Keep commit a9439a3, already on the branch. It adds the `PYTHONPATH` export that `run_review.sh:13` has and this script was missing, matching `main:20`. It is correct and its test passes.

For scope, the issue asks to move the review-decision logic (`should_review`, scope, fingerprint) into a single Python implementation so the shell and tests share one source of truth. Platform I/O, env-var defaults, output plumbing and the safety guard are not review-decision logic and should survive the migration unchanged. Please get `tests/test_artifact_paths.sh`, `tests/test_carry_forward_roundtrip.sh` and `tests/test_check_review_needed.sh` green, and confirm the 11 outputs above are still written to `$GITHUB_OUTPUT`.

joryirving (dismissed):
Request changes again, but the structural problems from the last round are all fixed and that work should stand. `OUTPUT_FILE` is back on `$GITHUB_OUTPUT`, the bogus `_lib.sh` source is gone and `platform_api.sh` and `artifact_paths.sh` are sourced properly, `assert_safe_artifact_paths` is called again, `REREVIEW_LABEL`, `platform_resolve` and `SKIP_IF_DIFF_UNCHANGED` are restored, and all four of `REPO`, `PR_NUMBER`, `COMMENT_MARKER` and `PUBLISH_MODE` are set before they are read. `tests/test_artifact_paths.sh` passes now. Do not undo any of that.

What remains is the migration's own decision logic, and it is wrong in a way the plumbing fixes cannot mask. Three suites fail on dd1a444.

`tests/test_check_review_needed.sh` reports 39 passed, 8 failed, and every failure is the same shape: the skip path never triggers.

```
FAIL: should_review=false when broad fingerprint matches (got 'true', expected 'false')
FAIL: should_review=false without force (matching fingerprint) (got 'true', expected 'false')
FAIL: should_review=false when everything matches (got 'true', expected 'false')
FAIL: should_review=false when a PR review fingerprint matches (review_verdict) (got 'true', expected 'false')
FAIL: should_review=false on matching fingerprint (got 'true', expected 'false')
FAIL: PR object is not fetched when skipping (got '1', expected '0')
FAIL: broad fingerprint has pipe delimiter (got 'no', expected 'yes')
FAIL: empty diff fingerprint contains 'empty-diff' (got 'no', expected 'yes')
```

The last two are the likely cause of the first five. `build_broad_fingerprint` is supposed to produce `<diff_fp>|cfg:<config_hash>`, and an empty diff is supposed to produce a fingerprint containing `empty-diff`. Neither marker is present in what the shell emits, so the value being compared against the stored `<!-- ai-pr-review-fingerprint:... -->` marker is not in the format that marker uses, and no comparison can ever match. Check what `precheck.py` writes into `BROAD_FINGERPRINT` and what the shell actually exports, since a fingerprint that never matches means the action re-reviews every push forever. The `PR object is not fetched when skipping` failure is the same bug seen from the other side: the skip short-circuit is never reached, so the fetch that should have been avoided still happens.

`tests/test_review_scope.sh` reports 18 passed, 3 failed:

```
FAIL: non-forced auto review uses incremental scope with a safe baseline (got 'full', expected 'incremental')
FAIL: safe incremental review trusts clean baseline (got 'false', expected 'true')
FAIL: non-forced dirty-baseline review remains incremental (got 'full', expected 'incremental')
```

`_detect_incremental_scope` is either not being reached or its result is not making it back into `effective_review_scope` and `baseline_clean`. Note the shell passes `ENABLE_INCREMENTAL_DETECTION`, `PREVIOUS_HEAD_SHA` and `PREVIOUS_BASE_SHA` into the Python; confirm all three arrive with the values the tests set up, and that the resulting scope is written to `$GITHUB_OUTPUT` rather than only into `precheck-result.json`.

`tests/test_carry_forward_roundtrip.sh` still fails, and this one now needs a decision rather than another deletion. The last round dropped `extract_review_metadata` without moving it anywhere. It is not in `precheck.py` on this branch, and `tests/test_carry_forward_roundtrip.sh` is not in the diff. That suite extracts the function out of `check_review_needed.sh` by regex at line 35 and sources it, so it is asserting on real code that no longer exists. Pick one and say which in the PR description:

either keep `extract_review_metadata` as a shell function in `check_review_needed.sh` so the roundtrip test keeps exercising the live implementation, or move its logic into `precheck.py` and rewrite `tests/test_carry_forward_roundtrip.sh` to call the Python directly. Both are acceptable. What is not acceptable is deleting it and leaving the test pointing at nothing, or re-adding a bash copy that exists only to satisfy the regex while the real work happens elsewhere.

The three acceptance criteria on #497 are about having exactly one implementation of scope detection and fingerprinting in the action path. That is satisfied by construction now. What is not yet satisfied is that the single implementation produces the same decisions the previous one did, which is what these 11 failures are measuring. Please get `tests/test_check_review_needed.sh`, `tests/test_review_scope.sh` and `tests/test_carry_forward_roundtrip.sh` green without weakening their assertions.

itsmiso-ai:
Pushed fix commit `699689d` to the existing PR branch. It completes the precheck migration without undoing the restored plumbing: `check_review_needed.sh` now calls `pr_reviewer.precheck` for marker-only decisions before the PR-object fetch, then for metadata-based scope/baseline after shell range-validation probes; carry-forward extraction is restored; `precheck-result.json` is allowlisted; and live-path unit coverage was added.

Verified locally with Python 3.14 / system PATH:
- `tests/test_check_review_needed.sh`: 47 passed
- `tests/test_review_scope.sh`: 21 passed
- `tests/test_carry_forward_roundtrip.sh`: 24 passed
- `tests/test_artifact_paths_drift.sh`: passed
- `tests/unit/test_precheck.py`: 114 passed
- `pytest tests/ --ignore=tests/test_tool_surface.py -q`: 1523 passed
- shellcheck all shell scripts: passed
- YAML/action metadata checks: passed

The full local suite is blocked only by the existing 1Password SSH-signing hook in `tests/test_tool_surface.py`; CI's fresh Ubuntu runner should not have that local credential-manager issue.

itsmiso-ai:
CI is green on the pushed head `699689d`: `validate` passed and the dogfooding `review` job passed. The action's fresh review is also now `APPROVED`. GitHub still reports the PR as `BLOCKED` because the older standing member `CHANGES_REQUESTED` review remains; that needs dismissal by the reviewer/maintainer (or the repository's review policy must be satisfied) before merge.

Issue #448, opened 2026-07-30
ai_fallback_base_url/api_key don't inherit from primary, unlike the smart-route inputs


joryirving:
## Inconsistency

`ai_smart_base_url`, `ai_smart_api_format`, and `ai_smart_api_key` all default to their `ai_*` primary equivalents when blank. The fallback inputs don't: `ai_fallback_base_url` and `ai_fallback_api_key` default to empty.

When the fallback model lives on the same gateway as the primary (a very common setup — one LiteLLM/Bifrost endpoint multiplexing providers), it's natural to set only `ai_fallback_model` and expect it to work like `ai_smart_model` does. Instead the fallback silently has no endpoint.

## Suggested fix

Either:
- Inherit blank fallback base_url/format/key from the primary, matching the smart-route behavior, or
- Fail fast at startup when `ai_fallback_model` is set but `ai_fallback_base_url` is blank, so the misconfiguration surfaces before a primary outage does.

The first keeps the "separate endpoint" use case working (explicit values still win) while making the same-gateway case zero-config.

PR #555, opened 2026-09-07
fix(config): let a single-provider configuration start

hampsterx:
Setting `ai_base_url` without `ai_fallback_model` fails configuration validation and exits 1 before any model call:

```
ERROR: AI_FALLBACK_MODEL is required when AI_FALLBACK_BASE_URL is set
```

`action.yml:840` fills the fallback base URL from the primary whenever the input is blank, so `AI_FALLBACK_BASE_URL` is non-empty in every otherwise valid single-provider configuration reaching `config.sh:318`. That check then rejects a non-empty base URL with no fallback model, so it fires for callers who configured no fallback at all. A workflow cannot opt out: the input defaults to empty and GitHub expressions treat an empty string as falsy, so `ai_fallback_base_url: ""` is indistinguishable from omitting it.

**This is a v2 regression.** `v1` binds `AI_FALLBACK_BASE_URL: ${{ inputs.ai_fallback_base_url }}` with no inheritance. The inheritance arrived in #458 (`e71b4a4`) resolving #448. The check itself is older than both, dating to the first commit (`89c8d70`) and moved verbatim into `config.sh` by the #324 split, so #458 added inheritance without reconciling a check already there.

The README quickstart is one of the affected callers: it pins `@v2` and sets `ai_base_url` with no `ai_fallback_*` inputs, as do both files in `examples/`. Nothing in CI exercises the path, because `.github/workflows/ai-pr-review.yaml` sets all four `ai_fallback_*` inputs explicitly.

**The fix restores work already done here.** `6a82b27`, an earlier head of #458 that was force-pushed away, carried exactly this exemption plus a shell test, and its commit message describes it: *"updated the validation ... to skip when the base_url was inherited from primary"*. This PR restores the comparison and adds a test covering it.

```diff
-if [[ -n "$AI_FALLBACK_BASE_URL" && -z "$AI_FALLBACK_MODEL" ]]; then
+if [[ -n "$AI_FALLBACK_BASE_URL" && -z "$AI_FALLBACK_MODEL" && "$AI_FALLBACK_BASE_URL" != "$AI_BASE_URL" ]]; then
```

Nothing in the resolved environment records whether a value was supplied or inherited, so the comparison is what separates the two. In effect `ai_fallback_model` alone gates the tier, which is the principle the smart route already uses. A base URL differing from the primary was chosen separately, so a missing model there is still an error.

**Alternative considered**: gating the inheritance in `action.yml` so the expression yields empty without a fallback model. It fixes direct `run_review.sh` callers too, which this doesn't. Against it: it reopens `test_action_inputs.py:193`, which pins the current expression as #448's regression test, and it puts tier-gating logic in a YAML expression where every other tier decision is made in `config.sh`. Happy to switch if you prefer that boundary.

**Out of scope, noted while tracing the newly reachable path**: `review.sh:402` tests `[[ "$ANALYSIS_ENGINE" == "$AI_FALLBACK_MODEL"* ]]`, which with an empty model is `== *`. The `[ -f ai-response.fallback.json ]` guard covers a clean workspace; a reused one could pick stale token-usage figures for the step summary. Pre-existing, affects summary reporting only, not left in this diff.

## Test plan

- [x] New `tests/test_fallback_config_validation.sh`: 4 cases, each asserting exit status *and* which validation produced it, since `config.sh` has earlier guards that also exit 1
- [x] Fails against `main` on exactly the inherited-URL case (`1|needs-model`), other three green
- [x] Deliberately broken fixture fails all four as `other-guard`, so a case cannot pass without reaching the check
- [x] Runs under `env -i`; verified identical under a hostile ambient environment and a token-free shell
- [x] Full bash suite green; `pytest` 1686 passed, coverage 80.95% vs the 72 gate
- [x] `shellcheck` 0.9.0 over all 55 shell files and `bash -n` clean


itsmiso-ai (approved):
LGTM — no blocking findings.

The validation change matches the actual action contract: when the fallback URL is only the inherited primary URL, `AI_FALLBACK_MODEL` remains the tier gate; when a distinct fallback endpoint is configured, the missing-model misconfiguration still fails closed.

Verified locally:
- new fallback validation suite: 4/4 passing on this branch
- same suite against `origin/main`: fails only on the inherited-URL/no-model regression case
- full bash suite: green
- Python suite: 1685 passed, 1 skipped
- `git diff --check`: clean

GitHub checks had not reported for the fork branch when I reviewed, so this approval is based on local verification.

Issue #409, opened 2026-07-11
fix: corpus.sh unconditionally emits Linked Issue Context / Evidence Providers headers, and evidence-providers.md still writes a 'not configured' fallback sentence


Tanguille:
## Observation

Follow-up to #399 / #400. The fix in #400 removed the fallback sentence from
`linked-issues.md`, but left the section header in `scripts/sections/corpus.sh`
unconditional:

```sh
# corpus.sh:110
echo "# Linked Issue Context"
cat linked-issues.md
```

This prints the header even when `linked-issues.md` is now empty (post-#400), so
the model still sees a section boundary with nothing under it — and, consistent
with the behavior #399 originally reported, still emits a "Linked Issue Fit"
section explaining that there's nothing to report, rather than omitting it.

Separately, and apparently missed by #400: `scripts/run_evidence_providers.py`
has the exact same bug #399 originally flagged, in a second code path that
was never touched:

```python
# run_evidence_providers.py:281-283
if not config_path_raw:
    write_outputs(summary, "No evidence providers configured.")
    return 0
```

`write_outputs` writes that sentence into `evidence-providers.md`, and
`corpus.sh:140` unconditionally headers it the same way:

```sh
# corpus.sh:140
echo "# Evidence Providers"
cat evidence-providers.md
```

## What we saw

PR review on a routine Renovate dependency bump (no linked issue, no evidence
providers configured in our workflow):

> ### Linked Issue Fit
> No linked issues were provided for this PR. As a routine dependency update, no specific issue acceptance criteria need to be verified.
>
> ### Evidence Provider Findings
> No evidence provider findings were generated.

Both sections should be entirely absent per `default_system_prompt.txt`'s own
stated conditional ("a Linked Issue Fit section when issue context is
available", "an Evidence Provider Findings section when provider output
exists").

## Suggested fix

- `corpus.sh`: only emit the `# Linked Issue Context` header when
  `linked-issues.md` is non-empty (mirrors what #400 already did for the file
  content — just needs to extend to the header line too).
- `run_evidence_providers.py`: leave `evidence-providers.md` empty instead of
  writing "No evidence providers configured." / "not found" fallback
  sentences when there's nothing to report — same treatment #400 gave
  `context.sh`.
- `corpus.sh`: only emit the `# Evidence Providers` header when
  `evidence-providers.md` is non-empty.

Happy to be told this is by design (e.g. corpus self-documentation) — but
since #399 already established the model reliably fills in any section it
[...]

Issue #101, opened 2026-06-03
fix: remove misleading “Tool harness disabled” text from planning corpus


joryirving:
## Problem

When `tool_mode=plan_execute_once` is enabled, `scripts/run_review.sh` builds the first `review-corpus.md` before running the tool harness. At that point, it creates a default `tool-harness.md` containing:

```md
Tool harness disabled.
```

That placeholder then appears in the exact corpus passed to the tool planner. So the planner is asked to plan tool usage while also seeing a `# Tool Harness Findings` section that says the harness is disabled.

This is confusing context for local/smaller models and can contribute to planner responses with zero tool calls or inconsistent tool-use behavior.

## Current flow

Relevant area:

- `scripts/run_review.sh`
  - creates default `tool-harness.md` before the initial corpus build
  - builds `review-corpus.md`
  - truncates it to `review-corpus.truncated.md`
  - then runs `scripts/run_tool_harness.py`
  - then rebuilds the corpus with real harness output

The first corpus is used by the tool planner, so the placeholder text matters.

## Desired behavior

The planning corpus should not say the tool harness is disabled when the harness is about to run.

Acceptable options:

1. Build a separate planning corpus that omits the `# Tool Harness Findings` section entirely.
2. Replace the placeholder with neutral text when `TOOL_MODE=plan_execute_once`, for example:

```md
Tool harness planning pending.
```

3. Move the default `Tool harness disabled.` placeholder so it is only used when `TOOL_MODE=off`.

## Acceptance criteria

- When `TOOL_MODE=plan_execute_once`, the corpus passed to the tool planner does **not** contain `Tool harness disabled.`
- When `TOOL_MODE=off`, the final review corpus can still include a clear disabled/skipped message.
- Add a regression test that proves the planning corpus does not include misleading disabled text when the harness is enabled.
- Existing smoke tests and parser/enforcement tests still pass.

## Notes

This is probably not the only reason tools sometimes do not run, but it is noisy context that weakens the planner. The planner should see crisp, contradiction-free instructions, especially for local models.

itsmiso-ai:
## Saffron backlog grooming note

Applied by Saffron groomer.

**Recommendation:** `ready`
**Lane:** `normal`
**Confidence:** `high`

**Reason:** Bounded shell script fix with clear acceptance criteria and test requirements.
### Summary
[...]

PR #560, opened 2026-09-08
fix(corpus): gate the Tool Harness Findings section on harness output

hampsterx:
## Problem

`README.md` already documents the intended behaviour:

> Every conditional review section is tied to a corpus trigger: **Linked Issue Fit** to
> linked-issue context, **Evidence Provider Findings** to provider output, **Tool Harness
> Findings** to tool output ... When a trigger is unmet the section is omitted entirely
> rather than emitted with "No findings" filler.

`default_system_prompt.txt` says the same thing, and is explicit about the failure mode:

> When a section's trigger condition is not met, omit the section entirely, emit no header,
> no placeholder sentence such as "No findings" or "None" [...] rather than including the
> section with an empty body.

With `tool_mode: off` there is no tool harness output, so the section should be absent. It
isn't. `scripts/sections/corpus.sh` emits the header with no presence gate, over a
`tool-harness.md` that the seeding block above it has already filled with the literal line
`Tool harness disabled.`

So the model receives a populated `# Tool Harness Findings` section and reports what it says.
**It is not confabulating.** This is the input-side twin of the #415 failure mode, with the
corpus supplying a *status* where the prompt's trigger expects *output*.

Observed across 12 consecutive reviews from a caller running with the harness disabled: the
section appeared in 11 of them, every time as `## Tool Harness Findings`, each carrying a
variant of the same sentence.

> The tool harness was disabled, so no automated tool results were available for this
> incremental review.

"for this incremental review" is the model echoing the `(incremental review)` suffix from the
`corpus.sh` header, which appears nowhere else in the pipeline. It is reading the section it
was handed.

Every neighbouring conditional section in the same block is gated:

| Section | Gate |
|---|---|
| Linked Issue Context | `[ -s linked-issues.md ]` |
| **Tool Harness Findings** | **none** |
| Evidence Providers | `[ -s evidence-providers.md ]` |
| CI Check Results | `[ -n "$CI_CHECKS_FILE" ] && [ -s ... ]` |

The comment three lines above the ungated block states the policy it is the exception to:

```sh
      # context.sh leaves linked-issues.md empty when there's no linked issue
      # (#399/#400) so the model sees no section boundary to react to. Gate
      # the header the same way, matching the CI Check Results pattern below.
```

The `strip_empty_conditional_sections.py` backstop has the matching gap: `SECTION_HEADINGS`
covers `linked_issue`, `evidence_provider` and `standards`, and `publish_helpers.sh` computes
those three presence signals and no fourth. So neither the input gate nor the output backstop
catches this section.

## A prior decision this touches

#101 fixed this text for the *planning* corpus, and #108's acceptance criteria explicitly kept
it in the review corpus:

> When `TOOL_MODE=off`, the final review corpus can still include a clear disabled/skipped
> message.

That predates #409/#415, which established the opposite policy ("empty means omit entirely,
header included") for every sibling section, and it predates the README paragraph quoted at
the top. This looks like an unreconciled older decision rather than an oversight, so it may be
worth an explicit call rather than a silent fix. Entirely possible the #108 criterion still
stands and this should be closed.

## Changes

- **`corpus.sh`**: gate the header on `[ -s tool-harness.md ]`. The three stubs that carry
  text keep their section, because the gate is on the file and not on `TOOL_MODE`: the
  `native_loop` planning placeholder from #101/#108, the cross-repository skip, and the
  harness failure notice.
- **`corpus.sh` seeding**: `off` truncates rather than testing for existence, so a file left
  in a reused workspace cannot present itself as this review's findings, the same reason
  `standards-context.md` is truncated. `native_loop` seeds on missing *or empty*, because
  `off` now leaves an empty file behind and testing `-f` alone would strand a later
  `native_loop` run in the same workspace without its planning marker, and the verdict turn
  with no section to substitute into.
- **`strip_empty_conditional_sections.py`** gains the tool-harness section keys, and
  **`publish_helpers.sh`** computes `TOOL_HARNESS_PRESENT` beside the existing three.

The new keys match the **whole heading**, unlike their three siblings. A prefix is safe for
`standards` (nothing but its own section is headed "Standards ...") and unsafe here, because
"Tool Harness ..." is a plausible change-by-change finding heading on a PR that touches
harness code while the reviewing action itself runs with tools off. Prefix matching deletes
`### Tool Harness Results Leak Secrets` along with its body. For the same reason the
`(incremental review)` qualifier is an allowlist rather than "any trailing parenthetical",
which would eat `### Tool Harness Results (Fork Skip)`. Leaving filler behind costs a few
words; deleting a finding costs the finding.

## Tests

- `tests/test_tool_harness_presence_signal.sh` is new, mirroring
  `test_standards_presence_signal.sh`. It lifts the seeding block out of `corpus.sh` by regex
  so a refactor fails the extraction loudly rather than passing vacuously.
- `tests/test_strip_empty_conditional_sections.py` gains tool-harness cases alongside the
  other three keys, including the CLI env path and preservation cases for findings whose
  headings begin with a section title or carry their own parenthetical.
- `tests/smoke_test.sh`: the #101 regression test asserted the off-mode text it is no longer
  correct to write. It now pins the empty file and both gate directions.
- `tests/test_tool_loop_wiring.sh`: one assertion pinned this block's *indentation* as a proxy
  for "the pending stub covers native_loop". The block is no longer nested, so it now pins the
  stub's text; the arm's behaviour is pinned by the new presence suite.

## Test plan

- [x] `pytest tests/` - 1704 passed
- [x] All `tests/test_*.sh` plus `tests/smoke_test.sh` green
- [x] Mutation-checked: reverting the header gate, the off-mode truncate, the missing-or-empty
      seed, the presence signal, the whole-title match, or the qualifier allowlist each turns
      a suite red
- [ ] CI on this PR


joryirving:
I reviewed the implementation, tests, and the decisions in #101/#108 and #409/#410/#415/#416. The central policy decision looks right: the newer explicit “omit conditional sections when their trigger is absent” contract should supersede #108’s older permissive statement that off mode *can* include a disabled message. #108 was primarily preventing contradictory planning context; it did not require disabled-mode filler permanently.

There is one correctness issue I think should be addressed before merge:

### P1 — `tool-harness.json` remains stale when `tool_mode=off`

This PR unconditionally truncates `tool-harness.md` in off mode to prevent stale findings on persistent/reused workspaces, but `tool-harness.json` is still initialized only when it does not exist:

```sh
if [ ! -f tool-harness.json ]; then
  # write off-mode JSON
fi
```

A previous native-loop run can therefore leave JSON state active during a later off-mode review. Downstream consumers read that file even when tools are off:

- `should_escalate()` can see stale `planning_error`/`error` and trigger `tool_planning_failed` when `escalate_on_tool_planning_failure=true`.
- Step-summary tool counts, success counts, cache-hit ratio, evidence digest, and tool-call summaries can report the preceding review’s harness run.
- Tool failure enforcement is gated off correctly, but escalation and telemetry are not equivalently gated.

That contradicts the reused-workspace rationale already applied to the Markdown artifact. I recommend overwriting both artifacts in off mode:

```sh
: > tool-harness.md
cat > tool-harness.json <<'EOF'
{"mode":"off","planned_request_count":0,"executed_request_count":0,"tool_results":[]}
EOF
```

Please add a regression test that seeds stale JSON containing an error, nonzero calls, usage, and evidence, switches to off mode, and verifies it is reset.

### P2 / optional hardening — deterministic stripping remains deliberately narrow

Whole-title matching is a sensible choice because deleting a real finding such as `### Tool Harness Results Leak Secrets` is worse than leaving a little filler. However, plausible section variants such as these will survive:

```md
## Tool Harness Findings — None
## Tool Harness Findings (disabled)
## Tool Harness Results: No tools enabled
## 🛠 Tool Harness Findings
```

I don’t consider this merge-blocking if the conservative false-negative tradeoff is intentional. A safer future refinement would make Tool Harness matching heading-level-aware: narrowly broaden recognized variants at `##`, while always preserving `###` and deeper change-specific headings. At minimum, tests or a comment should document which variants intentionally remain.

Otherwise the patch looks strong: corpus and publish gates share the same presence signal, `native_loop` correctly reinitializes an empty planning placeholder, real harness output is preserved, and the tests cover the highest-risk false-positive cases.

— Codex

joryirving (changes requested):
Requesting changes for the stale `tool-harness.json` issue described in my detailed review comment above. Off mode must reset the JSON artifact as well as the Markdown artifact so reused workspaces cannot leak prior harness errors, telemetry, calls, or evidence into escalation and reporting. The conservative heading-matching concern is optional hardening and is not part of this blocking request.

— Codex

hampsterx:
**Review actions on `aeff802`**

| Action | Disposition |
|---|---|
| P1, stale `tool-harness.json` in off mode | Fixed |
| P2, deterministic stripping remains narrow | Documented, not widened |

**P1.** You are right, and the inconsistency was mine: the reused-workspace
rationale was applied to the Markdown artifact and not to the JSON one, which is
the half that reaches further. Off mode now writes both. I confirmed the two
consumers you named before changing anything: `escalation.py` reads
`planning_error` and `error` out of that file, so a stale one escalates a review
that ran no tools, and the step summary reports its call counts, usage,
cache-hit ratio and evidence digest as though they belonged to the current
review.

The regression is seeded exactly as you asked, with an error, nonzero planned
and executed counts, a usage block and an evidence digest, and asserts off mode
clears all of it. I added its counterpart too, that `native_loop` leaves a real
harness JSON alone, since a reset that clobbered genuine output would pass the
first test on its own. Removing the fix turns four assertions red.

**P2.** The conservative tradeoff is intentional, so I have recorded it rather
than widened it. `test_documented_variants_deliberately_survive` pins your four
examples plus bare `## Tool Harness` and `## Tool Harness Summary` as known
false negatives. Heading-level awareness is the right shape if this is revisited;
what stopped me reaching for it now is that each earlier attempt to widen the
match re-opened the false-strip class, first as a bare prefix, then as a narrower
prefix that still ate `### Tool Harness Results Leak Secrets`, then as a trailing
parenthetical that ate `### Tool Harness Results (Fork Skip)`. Pinning the
survivors means a future widening has to argue with a test rather than a comment.

Full suite green: 1710 Python tests, all shell suites.

joryirving (approved):

PR #582, opened 2026-09-12
feat(change-anchors): extract deterministic change anchors from PR diffs

### LINKED ISSUE #571: [P1] Context: extract deterministic change anchors from PR diffs
## Summary

Add a pure-Python change-anchor extractor that turns the PR diff/file list into a small structured set of identifiers suitable for deterministic repository relationship scans.

This issue **only extracts anchors**. It must not perform repository-wide grep/search and must not wire anything into the model corpus yet.

## Why

The current repository-impact scan derives up to a small number of generic terms mainly from PR title/body/version hints. Those terms are useful for dependency/version enrichment, but they are not a reliable representation of the program elements actually being changed.

For repo-level review we want deterministic seeds such as:

- changed source files;
- functions/classes/types introduced or modified;
- imported/local modules touched by the diff;
- stable config/resource identifiers when obvious.

A follow-up can use these anchors to find callers, references and tests before the model starts exploring.

## Desired artifact

Produce a versioned JSON artifact, e.g. `change-anchors.json`:

```json
{
  "version": 1,
  "files": [
    {
      "path": "pr_reviewer/tool_executors.py",
      "language": "python",
      "symbols": [
        {"name": "git_grep", "kind": "function", "confidence": "high", "line": 123}
      ],
      "imports": ["pathlib", "pr_reviewer.platform"],
      "identifiers": []
    }
  ],
  "anchors": [
    {"value": "git_grep", "kind": "symbol", "source": "pr_reviewer/tool_executors.py", "confidence": "high"}
  ],
  "truncated": false
}
```

The exact schema can differ, but:

- version it;
- retain source file attribution;
- retain confidence/kind so later consumers can prefer high-signal anchors;
- cap everything deterministically.

## Input

Consume existing review artifacts rather than refetching GitHub data:

- `pr.diff` or `pr.diff.truncated` (prefer full local diff where safe; this is local parsing, not model context);
- `pr-files.json` / `pr-files.raw.json` for changed paths.

Do not call GitHub or the model.

## Extraction scope

Use lightweight regex/line heuristics only. **Do not add tree-sitter, language servers, or compiler dependencies in this issue.**

Initial language support should be intentionally small but useful:

### Python

Recognize changed/added declaration lines such as:

- `def name(`
- `async def name(`
- `class Name`

Recognize import/module lines when they appear in changed hunks:

- `import foo`
- `from foo.bar import Baz`

### JavaScript / TypeScript

Recognize common declarations in changed/added lines:

- `function name(`
- `class Name`
- exported function/class forms
- `const name = (` / arrow-function forms only when confidently distinguishable

Recognize ES imports/requires appearing in changed hunks.

### Go

Recognize:

- `func Name(`
- methods (`func (r Receiver) Name(`)
- `type Name struct/interface`

Recognize import paths appearing in changed hunks where practical.

### Generic fallback

For unsupported source languages, retain the changed file path/stem as an anchor. Do not harvest every token from changed lines; that recreates the noisy current term scan.

## Hunk/line handling

Use the unified diff to associate added/context lines with the correct file and, where practical, new-file line numbers.

Prefer anchors from **added or modified new-side code**. Deleted-only declarations should either:

- be emitted with a distinct kind/source-side marker; or
- be omitted initially.

Whichever behavior is chosen, document and test it.

Do not accidentally treat diff metadata (`+++`, `@@`, paths) as source declarations.

## Caps / noise control

Suggested defaults:

- max changed source files: 100 (align with existing PR file list cap where possible);
- max symbols per file: 20;
- max imports per file: 20;
- max global anchors: 200.

Deduplicate case-sensitively for code symbols. Preserve deterministic ordering by changed-file order then line number/name.

Avoid anchors that are clearly low-value:

- one-character names;
- language keywords;
- generic punctuation/token fragments;
- hashes/digests/URLs already handled elsewhere.

## Safety

This parser consumes attacker-controlled diff text.

- Never execute parsed content.
- Avoid catastrophic-backtracking regexes.
- Bound input/loop behavior.
- Treat file paths and symbol strings as untrusted when later rendered; this issue can keep the output JSON-only.

## CLI

Expose importable functions plus a small CLI suitable for orchestration, for example:

```bash
python3 -m pr_reviewer.change_anchors \
  --diff pr.diff \
  --files pr-files.json \
  --output change-anchors.json
```

## Expected files

- `pr_reviewer/change_anchors.py`
- `tests/test_change_anchors.py`
- `AGENTS.md` — document the new artifact/module

Optional thin script wrapper only if project convention needs it.

Do **not** touch `corpus.sh`, `run_tool_harness.py`, or `action.yml` in this issue.

## Tests

Create compact unified-diff fixtures covering:

- Python function/class/import additions;
- TS/JS function/class/import additions;
- Go function/method/type/import additions;
- multiple files/hunks;
- renamed file metadata;
- deleted-only symbol behavior;
- declarations appearing only as unchanged context vs actually modified lines (define intended behavior);
- duplicate anchors;
- generic unsupported language fallback;
- malformed/truncated diff;
- hostile/very long source lines;
- caps/truncation metadata;
- deterministic ordering.

## Acceptance criteria

- [ ] Diff + changed-file artifacts produce a versioned `change-anchors.json`.
- [ ] Python, JS/TS and Go common declarations are extracted with tests.
- [ ] Unsupported languages still contribute safe file/path anchors.
- [ ] Output is deterministic, deduplicated and capped.
- [ ] Parser executes no repository code and performs no network/model calls.
- [ ] Malformed/truncated diffs degrade cleanly rather than crashing the review pipeline.
- [ ] Focused tests cover false-positive boundaries and hostile input.

## Non-goals

- Full AST correctness.
- Caller/reference lookup.
- Test discovery.
- Reading entire changed files.
- Semantic/vector search.
- Corpus/prompt integration.

itsmiso-ai:
## What This Does

Implements the deterministic, bounded change-anchor artifact requested by #571.

- Adds `pr_reviewer.change_anchors`, an importable pure-Python extractor and CLI that turns `pr.diff` plus `pr-files.json` into versioned `change-anchors.json`.
- Extracts additions-only Python, JavaScript/TypeScript, and Go declarations/imports; unsupported source files produce a safe file-path anchor without token harvesting.
- Preserves deterministic changed-file/line/name ordering; applies explicit file, symbol, import, global-anchor, and input-byte caps with visible truncation metadata.
- Parses regular and C-style quoted Git diff paths (spaces, tabs, quotes, backslashes), rename metadata, malformed/truncated diffs, deleted files, and binary markers without executing parsed content.
- Keeps the module intentionally out of corpus/prompt wiring, tree-sitter/AST work, repository scanning, model calls, and network access - all are out of scope for #571.

## Safety

- The parser treats diff text and path strings as untrusted data and executes no repository code.
- CLI output is constrained to `--workspace-root` (default: `$GITHUB_WORKSPACE`, otherwise cwd), rejecting traversal, sibling-prefix escapes, NUL bytes, and symlink escapes.
- Tests cover malformed/hostile input, long lines, caps, path spaces/quotes, NUL output paths, traversal, and symlink output escapes.

## Validation

- `python3 -m pytest -q tests/test_change_anchors.py` - 65 passed locally.
- The previous CI failure was a test-environment bug: `GITHUB_WORKSPACE` overrode the subprocess cwd, causing the temporary test output to be rejected by the intended containment guard. The test now passes its temporary `--workspace-root` explicitly.

Fixes #571


joryirving (changes requested):
This is very close, but I found two correctness gaps that I think should be fixed before merging.

1. **Per-file cap truncation is not reflected in artifact metadata.** `_extract_file_anchors()` slices `fa.symbols` to `MAX_SYMBOLS_PER_FILE` and `fa.imports` to `MAX_IMPORTS_PER_FILE`, but that information never propagates back to `extract_change_anchors()`, so the top-level `truncated` flag remains `false` when those caps are the only truncation that occurred. That means an artifact can silently omit changed symbols/imports while claiming it is complete. #571 explicitly calls for deterministic caps and tests for caps/truncation metadata. Please propagate per-file truncation into the artifact-level flag (or add explicit per-file truncation metadata) and add regression tests for both symbol and import caps.

2. **Git diff paths containing spaces lose all diff-derived anchors.** `_DIFF_GIT_RE = r"^diff --git\s+(\S+)\s+(\S+)$"` only accepts whitespace-free paths. Git emits a normal path such as `foo bar.py` as `diff --git a/foo bar.py b/foo bar.py`; this header therefore fails to create `_FileState`. If `pr-files.json` is present the file still appears in the output, but it never gets enriched with the added lines from the diff, so changed functions/imports silently disappear. Please make diff path parsing handle normal Git path quoting/spacing semantics and add coverage for at least a Python filename containing spaces (and ideally quoted/escaped Git paths too).

Everything else I checked is nicely scoped to #571, the language heuristics are sensible for this first pass, and CI is green. Once those two silent-data-loss cases are covered, I’m happy with this.

itsmiso-ai:

**Total attempts:** 1669


**Attempts by lane:**
- **NORMAL:** 1669 attempts

**Final failure:** foreman fix exhausted 4/4 attempts on NORMAL (all coder tiers exhausted) (prfix-misospace-pr-reviewer-action-582)

**Last attempt:** I took another full pass and tested the parser against actual `git diff` output rather than only the synthetic fixtures. I found two correctness blockers that can silently lose or misattribute exactly the anchors #571 is meant to provide.

1. **Git's default octal path quoting is not decoded.** For a real Unicode filename such as `émoji.py`, default Git output is `diff --git "a/\303\251moji.py" "b/\303\251moji.py"`. `_parse_path_token()` and `_clean_diff_path()` only understand the named C-style escapes in `_PATH_ESCAPES`; an unknown escape consumes the backslash and keeps the next character, so `\303\251moji.py` becomes `303251moji.py` rather than `émoji.py`. `pr-files.json` will contain the actual Unicode filename, so the merge no longer recognizes the diff state as the same file: the literal-path file-list entry has no diff-derived symbols/imports, while a second bogus `303251moji.py` entry receives them. Under file caps those diff anchors can be dropped completely. Please implement Git-compatible octal escape decoding (including the UTF-8 byte sequence case) and add regression coverage using a Unicode filename / representative default-Git header.

2. **Multiline import state ignores unchanged new-side context.** `_FileState` records only `+` lines, then `_extract_file_anchors()` tries to infer Python `from ... import (` and Go `import (` block state from that additions-only stream. In the common case where the block already exists and the PR adds one member/path, the opening `from ... import (` / `import (` and closing `)` are context lines, so `in_from_block` / `in_block` never becomes true and the newly added import is silently missed. For example, a real Go diff for adding `"net/http"` to an existing `import (` block has only `+\t"net/http"`; the extractor returns no import anchor. Python has the same problem for an added name inside an existing parenthesized `from ... import (...)`. Please preserve enough new-side hunk/context structure to establish block state while still emitting anchors only from added lines, and add focused Python + Go regression tests.

The original cap/truncation and spaces/quoted-path fixes are good, and CI is green on this head. These two cases are the remaining blockers from my deeper pass.

**Latest note:** foreman fix exhausted 4/4 attempts on NORMAL (all coder tiers exhausted) (prfix-misospace-pr-reviewer-action-582)

Posted automatically by Dispatch on 2026-09-13T18:30:09.035Z

joryirving (approved):
Re-reviewed the current merged head against #571 and my previous request-changes findings. Both original correctness blockers are resolved: per-file symbol/import caps now propagate explicit truncation metadata to the artifact-level `truncated` flag, and diff path parsing now handles filenames with spaces plus C-style quoted/escaped Git paths, with focused regression coverage for both.

I also reviewed the subsequent CLI output-path containment changes and the later automated-review findings. The resolved-path containment correctly rejects traversal and symlink escapes; the redundant post-`resolve()` `is_symlink()` check is cleanup rather than a security hole. The CRLF concern is not actionable because `str.splitlines()` removes CR/LF line boundaries before parsing. The remaining synthetic NUL-filename concerns are not a merge blocker for a Git/GitHub-derived changed-file list and do not affect #571's parser contract.

Current CI is green. Approve.

joryirving (changes requested):
I took another full pass and tested the parser against actual `git diff` output rather than only the synthetic fixtures. I found two correctness blockers that can silently lose or misattribute exactly the anchors #571 is meant to provide.

1. **Git's default octal path quoting is not decoded.** For a real Unicode filename such as `émoji.py`, default Git output is `diff --git "a/\303\251moji.py" "b/\303\251moji.py"`. `_parse_path_token()` and `_clean_diff_path()` only understand the named C-style escapes in `_PATH_ESCAPES`; an unknown escape consumes the backslash and keeps the next character, so `\303\251moji.py` becomes `303251moji.py` rather than `émoji.py`. `pr-files.json` will contain the actual Unicode filename, so the merge no longer recognizes the diff state as the same file: the literal-path file-list entry has no diff-derived symbols/imports, while a second bogus `303251moji.py` entry receives them. Under file caps those diff anchors can be dropped completely. Please implement Git-compatible octal escape decoding (including the UTF-8 byte sequence case) and add regression coverage using a Unicode filename / representative default-Git header.

2. **Multiline import state ignores unchanged new-side context.** `_FileState` records only `+` lines, then `_extract_file_anchors()` tries to infer Python `from ... import (` and Go `import (` block state from that additions-only stream. In the common case where the block already exists and the PR adds one member/path, the opening `from ... import (` / `import (` and closing `)` are context lines, so `in_from_block` / `in_block` never becomes true and the newly added import is silently missed. For example, a real Go diff for adding `"net/http"` to an existing `import (` block has only `+\t"net/http"`; the extractor returns no import anchor. Python has the same problem for an added name inside an existing parenthesized `from ... import (...)`. Please preserve enough new-side hunk/context structure to establish block state while still emitting anchors only from added lines, and add focused Python + Go regression tests.

The original cap/truncation and spaces/quoted-path fixes are good, and CI is green on this head. These two cases are the remaining blockers from my deeper pass.

joryirving (approved):
Re-reviewed current head `0b8e555` against both prior review rounds. The blockers are resolved:

- per-file symbol/import caps now set explicit per-file truncation metadata and propagate to the artifact-level `truncated` flag;
- diff paths with spaces and C-style quoting are parsed without losing diff-derived anchors;
- Git's default octal path quoting is decoded byte-wise, including UTF-8 sequences such as `caf\303\251.py` → `café.py`, with regression coverage;
- Python `from ... import (` and Go `import (` block state now incorporates unchanged new-side context while still emitting anchors only from added lines, with focused regressions for both languages.

I also rechecked the surrounding parser/cap behavior rather than relying only on the automated review. The implementation remains scoped to #571, deterministic and bounded, and CI is green at this exact SHA. Approved.

PR #593, opened 2026-09-13
feat(tooling): add bounded find_files tool for filename/path discovery

### LINKED ISSUE #567: [P1] Tooling: add a bounded find_files tool for filename/path discovery
## Summary

Add a read-only `find_files` native-loop tool for locating repository files by filename/path pattern without requiring the reviewer to know an exact path first.

`list_tree` answers "what is around here?". `find_files` should answer "where is the config loader / auth middleware / matching test / package manifest?" without spending multiple tree-list calls.

This is intentionally a filename/path discovery primitive, not semantic code search.

## Desired contract

Suggested tool shape:

```json
{
  "pattern": "*config*",
  "path": ".",
  "max_results": 100
}
```

Semantics:

- `pattern` is required.
- Support a simple glob-style pattern against repo-relative paths or basenames. Prefer Python `pathlib` / `fnmatch` semantics over invoking a shell.
- `path` is optional and defaults to repository root.
- `max_results` defaults to `100`; clamp to a safe maximum such as `300`.
- Return sorted repo-relative file paths only. Directories may be omitted; `list_tree` covers directory discovery.
- Do not descend through `.git`.
- Do not follow symlinked directories outside the workspace.
- Do not return file contents.
- Keep the implementation deterministic and bounded.

Examples the tool should support:

```text
find_files(pattern="*config*")
find_files(pattern="test_*.py", path="tests")
find_files(pattern="*.toml")
find_files(pattern="*/route.ts")
```

If full `**` recursive glob semantics would complicate containment or portability, document and implement a deliberately smaller contract. The important part is that the schema and tests precisely define what matching means.

## Security requirements

This is model-controlled filesystem traversal over an untrusted checkout, so apply the same boundaries as `read_file` / the proposed `list_tree` tool:

- workspace containment;
- null-byte rejection;
- no traversal outside workspace;
- no symlink escape;
- no `.git` traversal;
- bounded result count;
- no execution of repository code;
- no raw shell expansion.

Prefer sharing existing path-resolution helpers instead of creating a second containment implementation.

## Expected files

Likely production files:

- `pr_reviewer/conversation.py`
- `scripts/run_tool_harness.py` if dispatch/argument normalization enumerates built-ins
- `README.md` if the public native-loop tool list is documented there
- `AGENTS.md` if the architecture/tool catalogue is enumerated there

Tests:

- add/extend `tests/test_tool_executors.py`
- extend `tests/test_run_native_loop_wiring.py`

## Test cases

Cover at least:

- exact-ish basename match;
- wildcard filename match;
- nested path match;
- scoped search under `path`;
- deterministic ordering;
- result cap;
- no matches returns an empty result, not an error;
- invalid/nonexistent root path returns a clean error;
- `../` path escape rejected;
- symlink escape rejected/not followed;
- `.git` files are not returned;
- schema advertisement in OpenAI/Anthropic payloads;
- end-to-end native-loop dispatch to the executor.

## Acceptance criteria

- [ ] Reviewer can locate files by path/name pattern without first discovering an exact path.
- [ ] Matching semantics are documented and pinned by tests.
- [ ] Results are stable, sorted, and capped.
- [ ] Filesystem traversal stays inside the workspace and excludes `.git`.
- [ ] No shell command/glob expansion is introduced.
- [ ] Tool schema and executor stay in sync.
- [ ] Existing tool harness tests remain green.

## Non-goals

- Searching file contents. `git_grep` covers that and will get a separate path-scoping improvement.
- Fuzzy or embedding-based search.
- Remote repository discovery.
- Building the deterministic repository map artifact.

itsmiso-ai:
Added a bounded find_files tool for filename/path discovery, complete with security guards and schema support for both OpenAI and Anthropic.

Fixes #567

_Opened by foreman on review GO (workload wl-misospace-pr-reviewer-action-567)._


itsmiso-ai:

**Total attempts:** 246


**Attempts by lane:**
- **NORMAL:** 246 attempts

**Final failure:** foreman fix exhausted 4/4 attempts on NORMAL (all coder tiers exhausted) (prfix-misospace-pr-reviewer-action-593)


**Failing run(s):**
- https://github.com/misospace/pr-reviewer-action/actions/runs/34734616879/job/103663656391

**Last attempt:** Two correctness issues to fix before merge:

1. **Directly-scoped `.git` traversal bypasses the exclusion.** `find_files()` prunes child directories named `.git`, but `_resolve_workspace_path()` accepts `.git` itself (and subpaths such as `.git/objects`). A call like `find_files(pattern="*", path=".git")` therefore starts `os.walk()` inside Git metadata and returns/traverses it. That violates #567's explicit `Do not descend through .git` / `.git files are not returned` contract. Please reject any resolved search root whose repo-relative path contains a `.git` component, and add regression coverage for `path=".git"` plus a nested `.git` root if useful.

2. **`max_results=0` does not follow the advertised clamp contract through native-loop dispatch.** `execute_tool_request()` currently passes `_opt_int(args.get("max_results")) or FIND_FILES_DEFAULT_MAX`; `_opt_int(0)` is `0`, so this becomes `100` instead of reaching `find_files()` and being clamped to `1`. The tool docs say values are clamped to `1..300`. Preserve `None` vs `0` explicitly and add a dispatch-path test for zero.

Otherwise this looks solid: the shared containment guard, symlink handling, sort-before-cap fix, schema wiring, and test coverage are all good, and CI is green on the current head.

**Latest note:** foreman fix exhausted 4/4 attempts on NORMAL (all coder tiers exhausted) (prfix-misospace-pr-reviewer-action-593)

Posted automatically by Dispatch on 2026-09-13T15:15:04.626Z

joryirving (changes requested):
Two correctness issues to fix before merge:

1. **Directly-scoped `.git` traversal bypasses the exclusion.** `find_files()` prunes child directories named `.git`, but `_resolve_workspace_path()` accepts `.git` itself (and subpaths such as `.git/objects`). A call like `find_files(pattern="*", path=".git")` therefore starts `os.walk()` inside Git metadata and returns/traverses it. That violates #567's explicit `Do not descend through .git` / `.git files are not returned` contract. Please reject any resolved search root whose repo-relative path contains a `.git` component, and add regression coverage for `path=".git"` plus a nested `.git` root if useful.

2. **`max_results=0` does not follow the advertised clamp contract through native-loop dispatch.** `execute_tool_request()` currently passes `_opt_int(args.get("max_results")) or FIND_FILES_DEFAULT_MAX`; `_opt_int(0)` is `0`, so this becomes `100` instead of reaching `find_files()` and being clamped to `1`. The tool docs say values are clamped to `1..300`. Preserve `None` vs `0` explicitly and add a dispatch-path test for zero.

Otherwise this looks solid: the shared containment guard, symlink handling, sort-before-cap fix, schema wiring, and test coverage are all good, and CI is green on the current head.

joryirving (changes requested):
One blocker remains from my previous review.

The `max_results: 0` dispatch bug is fixed correctly now: explicit zero survives argument normalization and `find_files()` clamps it to 1, with a focused regression test. CI is green on this head.

However, direct `.git` scoping is still possible. `find_files(path=".git", pattern="*")` passes through `_resolve_workspace_path()` because `.git` itself is inside the workspace and is not otherwise denied. `os.walk(resolved_root, followlinks=False)` then starts *inside* `.git`; `dirnames[:] = [d for d in dirnames if d != ".git"]` only prevents descending into `.git` when it appears as a child directory, so it cannot protect the starting root. The same applies to scopes such as `.git/objects`.

That violates #567's explicit acceptance/security requirement that `.git` traversal be excluded. Please reject a resolved search root that is `.git` or lies beneath `.git` (preferably by repo-relative path components rather than a string substring), and add direct executor/dispatch tests for `path=".git"` and a nested `.git/...` scope. After that I expect this to be ready to approve.

joryirving (approved):
Re-reviewed the follow-up fixes. Both previously blocking issues are now resolved:

- `max_results: 0` is preserved through dispatch and clamped by `find_files()` to 1, with a regression test.
- searches rooted at `.git` or any descendant such as `.git/objects` are rejected before `os.walk()` starts, closing the root-scope bypass; the new test covers both cases.

I also checked the latest delta specifically for collateral changes: it is limited to the `.git` root guard and its focused regression test. No new correctness or maintainability concerns from me.

CI is still queued/running on this head, so that should remain a merge gate, but the code review is clean. Approving.

Issue #570, opened 2026-09-12
[P1] Context: wire repository map into the review corpus and native-loop opening context


joryirving:
## Depends on

- #569 — deterministic bounded repository-map artifact

Do not start this until #569's JSON/Markdown contract is merged. This issue should consume that contract, not redesign it.

## Summary

Generate the repository map during a real review and make the compact Markdown map available to both:

1. corpus-only/single-shot reviews; and
2. the **first native-loop model turn**, before the model decides which tools to call.

The second requirement is the important one. A repo map that only reaches the final verdict after tool planning does not fix the discovery problem.

## Desired behavior

Every normal review should be able to begin with a bounded structural orientation such as:

```markdown
# Repository Map

- tracked files: 412
- primary languages: Python 52%, Shell 18%, YAML 14%
- major roots: `pr_reviewer/`, `scripts/`, `tests/`, `.github/`
- tests: `tests/` (168 files)
- workflows: `.github/workflows/`
- standards: `AGENTS.md`, `.github/ai-review-rules.md`
- important manifests: `action.yml`, `requirements.txt`
...
```

This is context, not evidence of correctness. The prompt should frame it as an orientation aid whose filenames/structure are untrusted repository data.

## Inputs

Add a small public configuration surface:

```yaml
repo_map_context:
  default: "true"
repo_map_max_bytes:
  default: "12000"
```

Suggested semantics:

- `repo_map_context=false` disables generation/injection entirely.
- `repo_map_max_bytes` bounds the Markdown artifact that reaches model context. Clamp/validate it using the project's existing numeric-input conventions.
- The JSON artifact may remain richer for diagnostics/tests, but model-facing text must respect the byte cap.

If maintainers prefer a different input prefix, keep the same two concepts: enable/disable + model-facing byte budget.

## Pipeline placement

Generate `repo-map.json` / `repo-map.md` after checkout/context initialization but before `build_review_corpus` and before native-loop planning.

Do not make map generation depend on linked-source enrichment, image analysis, evidence providers, or the model.

Generation failure should be **advisory**, not fatal:

- log a warning;
- leave the corpus section absent/empty;
- continue review;
- never invent a "map unavailable" section that the model then treats as findings.

## Corpus integration

Add a conditional section:

```markdown
# Repository Map
<bounded repo-map Markdown>
```

Suggested ordering:

1. PR Metadata / Classification
2. **Repository Map**
3. Incremental/linked issue context
4. PR files/diff
[...]

PR #595, opened 2026-09-13
feat(repo-map): add deterministic bounded repository-map builder

### LINKED ISSUE #569: [P1] Context: build a deterministic bounded repository-map artifact
## Summary

Add a pure-Python repository-map builder that produces a compact, deterministic structural summary of the checked-out repository.

This issue builds the artifact only. **Do not wire it into the review corpus or native-loop prompt yet.** Keeping generation separate makes the behaviour easy to unit-test and lets the follow-up integration PR deal with context budgeting independently.

## Problem

Today the reviewer gets excellent PR-local evidence but weak up-front repository orientation. The existing repository impact scan is seeded from a small set of words extracted from PR prose/version hints. Native tools can inspect arbitrary files, but the model first has to know what paths/subsystems exist.

We need a cheap structural map that can be generated on every review without running repository code or requiring language servers/embeddings.

## Desired outputs

Create a module that can emit both:

- machine-readable JSON, e.g. `repo-map.json`;
- compact Markdown, e.g. `repo-map.md`.

Suggested JSON shape (adjust as needed, but keep it documented/stable):

```json
{
  "version": 1,
  "summary": {
    "tracked_files": 1234,
    "directories": 85,
    "languages": {"Python": 52, "Shell": 14, "YAML": 18}
  },
  "roots": [
    {"path": "pr_reviewer", "files": 18},
    {"path": "scripts", "files": 27},
    {"path": "tests", "files": 90}
  ],
  "important_files": {
    "manifests": ["pyproject.toml", "requirements.txt"],
    "standards": ["AGENTS.md", ".github/ai-review-rules.md"],
    "workflows": [".github/workflows/ci.yaml"],
    "entrypoints": ["action.yml"]
  },
  "tree": ["...bounded structural entries..."]
}
```

The exact schema can differ, but it should be intentionally versioned because later code will consume it.

## Data source

Prefer **tracked files** from Git rather than unrestricted filesystem walking so generated artifacts, virtualenvs, caches and checkout-local junk do not pollute the map.

A safe approach is an argv-only call equivalent to:

```text
git ls-files -z
```

Do not invoke shell pipelines. Parse the result in Python.

If Git metadata is unavailable, fail cleanly or provide a documented bounded filesystem fallback. Do not silently emit a misleading partial map.

## Classification heuristics

Keep this intentionally lightweight. No AST parser or new heavy dependency in this issue.

Useful deterministic categories include:

- source-code extensions/languages;
- test paths (`tests/`, `test/`, `*_test.*`, `test_*.*`, etc.);
- package/dependency manifests (`pyproject.toml`, `requirements*.txt`, `package.json`, lockfiles, `go.mod`, `Cargo.toml`, etc.);
- CI/workflows (`.github/workflows`, Forgejo equivalents if obvious from existing project support);
- repository standards (`AGENTS.md`, `CLAUDE.md`, `.github/ai-review-rules.*`);
- common entry/config files (`action.yml`, Dockerfiles, Makefiles, compose files);
- migration/schema directories/files;
- API/controller/route directory-name hints;
- auth/security directory/file-name hints.

The map should report structure, not make risk judgments. Risk classification remains `classifier.py`'s job.

## Bounded output

The generator must stay useful on large repos.

Suggested controls exposed by the Python API/CLI:

- `max_depth` (default 3);
- `max_entries` (default ~500);
- `max_files_per_category` (default ~50);
- optional `max_markdown_bytes` cap.

Truncation must be explicit in the output (`truncated: true`, counts of omitted entries, or a Markdown note). Never silently drop entries while pretending the map is complete.

Stable ordering is required for repeatability and prompt-cache friendliness.

## Security / safety

- Do not read file contents in this issue; paths/metadata only.
- Do not execute repository code.
- Use argv-only Git subprocesses with a timeout.
- Treat filenames as untrusted text when rendering Markdown: do not let a filename inject arbitrary Markdown headings/fences into later prompts. Escape or render filenames in code spans safely.
- Handle unusual filenames, spaces, Unicode and newlines deliberately. `git ls-files -z` is preferred specifically because newline-delimited parsing is unsafe for arbitrary Git paths.

## CLI

Provide a small CLI so shell orchestration can consume it later, for example:

```bash
python3 -m pr_reviewer.repo_map \
  --workspace "$GITHUB_WORKSPACE" \
  --json repo-map.json \
  --markdown repo-map.md
```

If the repo convention strongly prefers a thin `scripts/build_repo_map.py` wrapper instead, keep core logic in the importable module and the wrapper trivial.

## Expected files

- `pr_reviewer/repo_map.py` — core implementation and rendering
- optionally `scripts/build_repo_map.py` — thin CLI wrapper only if needed by project convention
- `tests/test_repo_map.py`
- `AGENTS.md` — document the module/artifact contract

Avoid touching `scripts/sections/corpus.sh`, `action.yml`, or the default system prompt in this issue.

## Test cases

Use temporary Git repos to cover:

- basic nested project layout;
- deterministic ordering;
- tracked vs untracked files;
- hidden files;
- spaces/Unicode in filenames;
- newline in a tracked filename if feasible;
- depth cap;
- entry cap and explicit truncation metadata;
- category caps;
- language/extension counts;
- expected detection of tests/manifests/workflows/standards;
- empty repository;
- Git command failure/timeout;
- Markdown-safe rendering of hostile filenames such as a path containing backticks or heading-like text.

## Acceptance criteria

- [ ] `repo_map` can generate JSON + Markdown from a checkout without reading source contents.
- [ ] Output schema is versioned/documented.
- [ ] Output is deterministic across repeated runs on the same tree.
- [ ] Large repos are bounded by explicit caps with visible truncation metadata.
- [ ] Only tracked files influence the normal map.
- [ ] Arbitrary Git filenames cannot break the Markdown rendering contract.
- [ ] Generator executes no repository code.
- [ ] Focused unit tests cover normal, truncation and adversarial filename cases.

## Non-goals

- Adding the map to the review corpus.
- Symbol/import/caller analysis.
- Reading source contents.
- Embeddings/vector indexes.
- Persisting an index between workflow runs.
- Remote repository maps.

itsmiso-ai:
Updated test_important_files_cap to truly isolate the important-files cap as the PR reviewer asked: 60 requirements-NN.txt files in a single directory (only the manifests bucket exceeds the cap, roots/categories/entries/depth stay inside their limits). All 1746 tests pass.

Fixes #569


joryirving (changes requested):
Overall this is a strong implementation and the architecture is the right shape, but I found a few correctness issues in the artifact contract that should be fixed before #570 starts consuming it.

### 1. `important_files` truncation is currently silent

`build_repo_map()` caps each `important_files` bucket with `hits[:category_cap]`, but unlike `categories` and `roots`, it never records how many important-file hits were omitted and never adds a truncation reason.

That means a repo with, for example, >50 workflows/manifests/standards can return only the first 50 while `truncation.truncated` remains `false` as long as the tree/categories/roots happen to fit. That directly violates #569's requirement that truncation be explicit and never silently drop entries.

Please account for omissions from `important_files` in the truncation metadata (either a dedicated count/reason or a clearly documented shared count) and add a regression test where only an important-files bucket exceeds the cap.

### 2. `_code_span()` can be escaped by a filename containing exactly two consecutive backticks

The renderer switches to a fixed double-backtick delimiter whenever a path contains any backtick:

```python
if "`" in text:
    return "`` " + text + " ``"
```

That protects names containing a single backtick, and the current hostile fixture containing four backticks does not close a two-backtick span. But an untrusted path containing an exact run of two backticks *does* provide a matching delimiter and can terminate the span early. This matters outside the fenced tree too: roots, standards, manifests, categories, etc. are rendered as normal Markdown code spans.

Please choose a delimiter longer than the longest consecutive backtick run in the rendered path (the usual Markdown code-span strategy), and add a hostile fixture with an exact `` run, ideally in a path that appears in one of the non-tree sections such as `something``evil/AGENTS.md`.

### 3. `max_markdown_bytes` is not actually a hard maximum

`render_repo_map_markdown()` first fits content inside the cap, then appends the closing fence/truncation note afterward. For small or tight caps the returned UTF-8 document therefore exceeds `max_markdown_bytes`, despite the option being described as a byte cap. #570 is explicitly going to use this output as model-facing bounded context, so the size contract should be unambiguous before integration.

Please either make the returned document obey the configured maximum (including fence closure/note) or change the API/contract so callers cannot mistake it for a hard maximum. Given the issue wording, I strongly prefer enforcing the hard cap and adding `len(rendered.encode("utf-8")) <= cap` coverage for realistic caps plus a tiny-cap edge case.

Everything else I checked looks good: tracked-only `git ls-files -z`, deterministic ordering, fail-cleanly behavior, pure classification, no corpus wiring, and the test coverage are all in good shape. CI is green. Once these bounded-output / hostile-filename edges are fixed, I expect this to be approvable.

joryirving (dismissed):
Re-review complete. The three artifact-contract blockers from my prior review are resolved on this head: important-files truncation is explicitly accounted for and isolated by regression coverage; code-span delimiters are longer than the longest backtick run in hostile filenames; and max_markdown_bytes is now a hard UTF-8 cap including tiny/fence-cut cases. CI is green.

One merge-time note: GitHub currently reports this PR as dirty/non-mergeable because main advanced with #593 and both branches touch AGENTS.md. That conflict is unrelated to the repo-map implementation reviewed here, but the branch will need rebasing/conflict resolution before merge.

PR #596, opened 2026-09-13
feat(git_grep): add path scoping and explicit result limits

### LINKED ISSUE #568: [P1] Tooling: add path scoping and explicit result limits to git_grep
## Summary

Extend the existing `git_grep` native-loop tool so the reviewer can search file contents within a specific subtree and control the result cap.

This should remain the action's simple literal content-search primitive. Do not replace it with embeddings or a new search service.

## Current behavior

`pr_reviewer/tool_executors.py::git_grep()` currently runs approximately:

```python
git grep -n -- <pattern> .
```

and returns the first 60 matched lines.

That works for small repos, but once the reviewer has discovered a relevant subsystem it cannot narrow a common identifier such as `Config`, `validate`, `token`, or `handler` to that subtree. Broad matches burn tool-response bytes and make weaker/local models work harder to identify the useful references.

## Desired tool contract

Preserve the existing required `pattern` argument and add optional arguments:

```json
{
  "pattern": "load_config",
  "path": "pr_reviewer",
  "max_results": 40
}
```

Semantics:

- `pattern` remains required and literal.
- `path` defaults to `.`.
- `max_results` defaults to the current behaviour (`60`) so callers that do not use the new arguments are behavior-compatible.
- Clamp `max_results` to a bounded range, e.g. `1..200`.
- Path must be repository-relative and pass the same containment/sensitive-path validation used by content-bearing filesystem tools.
- Invoke git with `--` before the pathspec so the model cannot turn a path into an option.
- Preserve return shape compatibility where practical (`{"matches": [...]}`).

Example command construction:

```text
git grep -n -- PATTERN -- PATHSPEC
```

Use an argv list, never a shell command string.

## Important compatibility detail

Do not accidentally reinterpret `pattern` as a regex if the current tool/prompt contract describes it as literal. If the existing `git grep` invocation currently uses grep's default regex semantics, either:

1. preserve that exact behaviour and correct the tool description if it incorrectly says literal; or
2. make literal matching explicit with `-F` only if tests prove this is the intended existing contract.

Avoid silently changing match semantics in the same PR as path scoping.

## Security requirements

- Validate `path` through the shared workspace resolver.
- Reject traversal/symlink escapes.
- Apply sensitive-path restrictions consistently with `read_file` / `git_blame` if the search can reveal file content.
- Keep subprocess timeout handling.
- Keep result count bounded before returning output to the model.

## Expected files

- `pr_reviewer/conversation.py` — extend the advertised `git_grep` schema
- `scripts/run_tool_harness.py` — only if normalization/dispatch must explicitly accept the new args
- `README.md` / `AGENTS.md` only if they document the argument surface

Tests:

- extend an existing tool-executor test module or add `tests/test_tool_executors.py`
- extend `tests/test_run_native_loop_wiring.py` for argument forwarding

## Test cases

Cover:

- no `path` / no `max_results` preserves current behavior;
- scoped search returns only matches under the selected path;
- `max_results` below 60 works;
- overly large `max_results` is clamped;
- path beginning with `-` cannot become a git option;
- `../` traversal rejected;
- symlink escape rejected;
- sensitive-path policy is preserved;
- timeout/error handling remains clean;
- model-emitted optional args survive normalization and reach the executor.

## Acceptance criteria

- [ ] Existing `git_grep({pattern})` calls still work.
- [ ] Reviewer can restrict grep to a repository subtree.
- [ ] Reviewer can request a smaller bounded result set.
- [ ] Search result ordering remains deterministic enough for repeated runs/tests.
- [ ] No shell execution is introduced.
- [ ] Path containment/security tests cover adversarial arguments.
- [ ] Schema, normalization and executor agree on the optional arguments.
- [ ] Existing native-loop tests stay green.

## Non-goals

- Filename discovery (`find_files`).
- Tree discovery (`list_tree`).
- Symbol-aware AST references.
- Remote repository search.
- Semantic/vector search.

itsmiso-ai:
Extended `git_grep` to support optional `path` scoping and `max_results` capping, including updated schema, normalization, and security-conscious executor logic.

Fixes #568

_Opened by foreman on review GO (workload wl-misospace-pr-reviewer-action-568)._


joryirving (changes requested):
The path-scoping/result-limit work looks good overall, but there is one security-contract blocker in the `git_grep` result path.

### Broad scopes can still leak sensitive-file contents, and grep redaction is currently discarded

The new direct-path check correctly rejects `path=".env"`, but `_resolve_workspace_path()` only validates the requested scope itself. A broad scope such as `path="."`, `path="config"`, or the backward-compatible no-`path` call can still match sensitive descendants and return their contents.

Concrete case: a tracked `.env` containing `TOKEN=supersecret`, followed by `git_grep({"pattern": "supersecret", "path": "."})`. The workspace root passes the resolver, `git grep` returns the `.env` line, and that raw match reaches the model.

There is a second half to the same bug in `execute_tool_request()`: it does

```python
text = "\n".join(matches)
text, _ = mask_and_truncate(text, max_response_bytes)
tool_result["result"] = {"matches": matches[:max_results]}
```

so the sanitized/bounded `text` is thrown away and the original unredacted `matches` are returned. That means credential-like content in *any* matched file also bypasses the shared redaction helper, and `max_response_bytes` is not actually applied to the returned grep payload.

This conflicts with #568's explicit requirement to apply the sensitive-path policy consistently for a content-bearing search.

Please make the returned `matches` honor both boundaries: sensitive descendants must not expose their contents through broad scopes, and the payload actually returned to the model must use the redacted/bounded form rather than discarding it. Add regressions for at least:

- tracked `.env` + whole-repo / `path="."` grep does not expose its secret;
- a credential-like value in a normal searchable file is redacted in the returned grep result;
- the returned grep payload respects `max_response_bytes` (not only `max_results`).

Everything else I checked is in good shape: scoped argv is safe, the second `--` correctly protects dash-leading pathspecs, result caps clamp as intended, regex semantics are documented accurately, optional args survive normalization, and CI is green.

itsmiso-ai:

**Total attempts:** 367


**Attempts by lane:**
- **NORMAL:** 367 attempts

**Final failure:** foreman fix exhausted 4/4 attempts on NORMAL (all coder tiers exhausted) (prfix-misospace-pr-reviewer-action-596)

**Last attempt:** One security-boundary bug remains in the new `git grep -z` parsing, so I don't think this head should merge yet.

`git_grep()` currently does:

```py
lines = result.stdout.strip().splitlines()[:max_results]
return {"matches": [_redact_grep_match(line, workspace_root) for line in lines]}
```

The whole reason for `-z` is that Git prints pathnames verbatim and delimits them with NULs. Git permits newlines in tracked filenames, so splitting on newlines *before* consuming the NUL-delimited pathname breaks the record framing.

This is more than cosmetic because `_redact_grep_match()` is now a security boundary. A concrete adversarial case is a tracked sensitive filename whose name ends in a newline, e.g. `.env\n`, containing an arbitrary marker value that `mask_secrets()` does not recognize. Git's record is effectively:

```text
.env\n\0<line>\0<content>\n
```

`splitlines()` turns that into two pseudo-records. The second begins with `\0`, so `_redact_grep_match()` sees `path == ""`; `_resolve_workspace_path("")` resolves to the workspace root and succeeds, and the sensitive file content can then be returned as `:<line>:<content>`. The generic credential masker is only heuristic, so it does not restore the path-based guarantee.

Please parse the raw `-z` stream before any newline splitting: consume `path\0line-number\0content\n` records by locating the first and second NULs, then the record-ending newline. Apply `max_results` after records are parsed. Binary-match output can remain a separate case.

Please add a regression with a real tracked filename containing a newline, ideally a sensitive case such as `.env\n`, and use content that deliberately does *not* look like a token/key-value secret so the test proves the sensitive-path policy itself is doing the blocking. A normal newline-containing filename/provenance test would also be useful.

Everything else I checked on `069670ff` looks good: the earlier broad-scope `.env` leak is fixed for normal paths, returned normal-file credentials are now actually redacted, path scoping/argv handling is safe, result caps and normalization line up, and CI is green.

**Latest note:** foreman fix exhausted 4/4 attempts on NORMAL (all coder tiers exhausted) (prfix-misospace-pr-reviewer-action-596)

Posted automatically by Dispatch on 2026-09-14T17:45:11.087Z

joryirving (changes requested):
One security-boundary bug remains in the new `git grep -z` parsing, so I don't think this head should merge yet.

`git_grep()` currently does:

```py
lines = result.stdout.strip().splitlines()[:max_results]
return {"matches": [_redact_grep_match(line, workspace_root) for line in lines]}
```

The whole reason for `-z` is that Git prints pathnames verbatim and delimits them with NULs. Git permits newlines in tracked filenames, so splitting on newlines *before* consuming the NUL-delimited pathname breaks the record framing.

This is more than cosmetic because `_redact_grep_match()` is now a security boundary. A concrete adversarial case is a tracked sensitive filename whose name ends in a newline, e.g. `.env\n`, containing an arbitrary marker value that `mask_secrets()` does not recognize. Git's record is effectively:

```text
.env\n\0<line>\0<content>\n
```

`splitlines()` turns that into two pseudo-records. The second begins with `\0`, so `_redact_grep_match()` sees `path == ""`; `_resolve_workspace_path("")` resolves to the workspace root and succeeds, and the sensitive file content can then be returned as `:<line>:<content>`. The generic credential masker is only heuristic, so it does not restore the path-based guarantee.

Please parse the raw `-z` stream before any newline splitting: consume `path\0line-number\0content\n` records by locating the first and second NULs, then the record-ending newline. Apply `max_results` after records are parsed. Binary-match output can remain a separate case.

Please add a regression with a real tracked filename containing a newline, ideally a sensitive case such as `.env\n`, and use content that deliberately does *not* look like a token/key-value secret so the test proves the sensitive-path policy itself is doing the blocking. A normal newline-containing filename/provenance test would also be useful.

Everything else I checked on `069670ff` looks good: the earlier broad-scope `.env` leak is fixed for normal paths, returned normal-file credentials are now actually redacted, path scoping/argv handling is safe, result caps and normalization line up, and CI is green.

joryirving:
Fixed in ceefbd9.

The raw git grep -z stream is now consumed as path-NUL-lineno-NUL-content records before the result cap is applied. A newline inside a pathname therefore cannot split the path/content boundary before the sensitive-path check. Binary-match output remains a separate no-NUL record case.

Added real-Git regressions for:
- a force-tracked .env filename ending in a newline, with a bare non-credential marker value; the marker is path-redacted at both git_grep and executor boundaries
- a non-sensitive filename containing a newline, preserving its parsed path:lineno:content provenance

Verified: python3.14 -m pytest tests/ -v --tb=short — 1905 passed, 2 subtests passed.

joryirving (approved):

Issue #252, opened 2026-06-14
CI: catch the bug classes that mocks hide (golden fixtures, adversarial fences, actionlint, UA guard)


joryirving:
## Motivation

Several recent bugs passed green CI because the **tests encoded the same blind spot as the code**:
- Forgejo commit-status normalization read `state` while the real per-entry field is `status` (#244) — the mock used `state`, so it passed.
- The untrusted-data fence (#250) was escapable by content containing the closing delimiter — the hostile fixture omitted the delimiter, so it passed.
- The konflate evidence provider used urllib's default UA → Cloudflare BIC 403 — no test exercised UA/edge behavior.
- `PLATFORM:` landed inside a `run:` block instead of `env:` (#243) — no workflow lint.

Generic CI can't detect "the mock is wrong," so each class needs a targeted guard.

## Proposed (forge-agnostic, benefits all consumers)

1. **Golden fixtures recorded from real forges.** For `forgejo_backend`, capture real `/api/v1` responses (e.g. from a public Gitea/Forgejo instance) once, commit them, and parse against them — instead of hand-written mocks. Catches field-shape drift (`state` vs `status`). Highest value.
2. **Adversarial-fixture convention for security boundaries.** Any sanitizer/fence must have a test that feeds the boundary token / hostile delimiter itself (the #250 fix is the first example). Document it; apply to future fences.
3. **`actionlint` in CI.** Catches a broad class of workflow/`action.yml` shell + expression bugs.
4. **Outbound-HTTP User-Agent test.** Assert the action's HTTP helpers (`web_fetch`, `web_search`) set a non-default `User-Agent` (urllib's default trips CDN bot checks like Cloudflare BIC 1010). Cheap regression guard.

#3 and #4 are quick wins; #1 is the structural fix; #2 is mostly discipline.

PR #598, opened 2026-09-13
feat(tools): add bounded list_tree tool for repository discovery (#566)

### LINKED ISSUE #566: [P1] Tooling: add a bounded list_tree tool for repository discovery
## Summary

Add a read-only `list_tree` native-loop tool that lets the reviewer discover repository structure without already knowing exact paths.

This closes a basic discovery gap in the current tool surface: `read_file` requires a known path and `git_grep` requires a known search term. A reviewer dropped into an unfamiliar repo needs a safe way to ask "what exists here?" before it can choose the right files to inspect.

## Current behavior

The native loop currently exposes tools such as:

- `read_file`
- `git_grep`
- `git_log`
- `git_blame`
- `gh_api`
- `web_fetch`
- optional `web_search`
- fixed-catalog `run_command`

None provides directory/tree discovery.

## Desired tool contract

Add a tool named `list_tree` with arguments roughly:

```json
{
  "path": "optional/repo/relative/path",
  "depth": 2,
  "max_entries": 200
}
```

Suggested semantics:

- `path` defaults to repository root (`.`).
- `depth` defaults to `2`; clamp to `1..4`.
- `max_entries` defaults to `200`; clamp to a conservative upper bound such as `500`.
- Return deterministic, sorted repo-relative paths.
- Include enough type information to distinguish files from directories, e.g. `{path,type}` rows or a stable textual tree.
- Do not follow symlinks outside the workspace.
- Do not return sensitive-file contents. This tool lists names only, but it should still avoid traversing `.git` internals and other explicitly denied implementation directories if the existing path guards already establish such a policy.
- Respect the same workspace containment boundary used by `read_file`.
- Apply the harness's normal output-size truncation/capping.

The model must not be able to supply arbitrary shell commands. Implement this directly in Python using `pathlib` / filesystem traversal, not by adding an unrestricted `find` command to `run_command`.

## Security requirements

This tool becomes part of the prompt-injection security boundary because untrusted PR content can influence which arguments the model emits.

Required properties:

1. Resolve `path` through the existing workspace path guard (or a deliberately shared helper with identical containment/symlink rules).
2. Reject null bytes and paths escaping the workspace.
3. Do not follow symlinked directories outside the repository.
4. Do not descend into `.git`.
5. Keep traversal bounded by both depth and entry count.
6. Return data only; never execute repository code.

Prefer reusing `_resolve_workspace_path` from `pr_reviewer/tool_executors.py` instead of re-implementing containment logic.

## Implementation notes

The tool has three surfaces that must remain in sync:

1. executor implementation;
2. tool schema advertised to OpenAI/Anthropic models;
3. native-loop dispatch/normalization.

Do not add the executor without wiring all three.

Use stable ordering (`sorted`) so tests, prompt caching, and repeated reviews are deterministic.

## Expected files

Likely production files:

- `pr_reviewer/tool_executors.py` — `list_tree()` implementation
- `pr_reviewer/conversation.py` — `TOOL_SCHEMAS` entry
- `scripts/run_tool_harness.py` — dispatch/normalization wiring if the harness enumerates tool names/args there
- `AGENTS.md` — add the new built-in tool to architecture/tool-surface documentation if that list is maintained there
- `README.md` — update native-loop tool documentation if the public tool list is enumerated

Tests:

- add or extend `tests/test_tool_executors.py`
- extend `tests/test_run_native_loop_wiring.py` for schema advertisement + routing
- add any schema-equivalence/allowlist test that currently pins built-in tool names

## Test cases

At minimum cover:

- root listing at default depth;
- nested `path` listing;
- deterministic sort order;
- `depth` clamping;
- `max_entries` cap;
- nonexistent path;
- file passed as `path` (define one consistent behavior: either one file row or a clean error);
- `../` traversal attempt rejected;
- symlink inside repo pointing outside workspace rejected/not traversed;
- `.git` is not traversed;
- tool schema advertises the expected args;
- a native-loop model request for `list_tree` reaches the executor and folds the result back into the conversation.

## Acceptance criteria

- [ ] Native-loop models can call `list_tree` without knowing any repository paths in advance.
- [ ] Output is deterministic and bounded.
- [ ] Traversal cannot escape the checked-out workspace.
- [ ] `.git` internals are not exposed through traversal.
- [ ] No arbitrary command execution is introduced.
- [ ] OpenAI and Anthropic schema renderings both include the new tool through the existing shared catalogue.
- [ ] Native-loop routing executes the tool successfully.
- [ ] Security/adversarial tests cover traversal and symlink escape attempts.
- [ ] Existing tool-harness tests remain green.

## Non-goals

- Semantic code search.
- Remote/related repository trees.
- Building the persistent repository map artifact.
- Reading file contents.

Those are separate follow-ups so this remains a small, reviewable primitive.

joryirving:
## Summary

Closes #566. Adds a read-only `list_tree` native-loop tool that lets the reviewer answer "what is around here?" in an unfamiliar repo — the structural discovery primitive that `read_file` (needs exact path) and `git_grep` (needs exact term) both lack.

## What's included

- **Executor** (`pr_reviewer/tool_executors.py`): bounded top-down walk (depth 1..4, max_entries 1..500), sorted at each level, `.git` never traversed, symlinks skipped entirely (consistent with `find_files`), returns `{path, type}` rows with `total` + `truncated` flag. Path goes through `_resolve_workspace_path` (null bytes, escapes, sensitive patterns rejected). A file passed as `path` returns a one-row listing.
- **Dispatch**: `execute_tool_request` arm with row-boundary byte truncation (never splits an entry at a byte cut).
- **Schema** (`conversation.py`): `TOOL_SCHEMAS` entry (OpenAI/Anthropic), all params optional.
- **Harness**: `import list_tree` from tool_executors.
- **Docs**: AGENTS.md (tool_executors bullet), README.md (tool list), SECURITY.md allowlist (added `find_files` + `git_log`/`git_blame` which were stale).
- **Tests**: 18-test executor battery mirroring `find_files` (deterministic order, depth/clamp bounds, `.git`/symlink guards, file-as-path single row, hostile names with newline per #252), 2 wiring tests (schema advertised both formats + dispatch end-to-end), pinned catalogue updated.

## Test plan

- [x] Local pytest suite green (1613 passed, 1 pre-existing env error in `test_tool_surface.py::test_git_log_returns_history` due to 1Password GPG signing — verified pre-existing on main)
- [ ] CI green
- [ ] PR Reviewer action approves

## Notes

- The WIP branch (`origin/foreman/wl-misospace-pr-reviewer-action-566/issue-566`) had a partial executor that listed symlinked directories as "file" type (misleading, inconsistent with `find_files`). This PR fixes that and completes the full wiring (schema, tests, docs).
- The clamping logic distinguishes "unset" (None → default) from an explicit out-of-range value (0 → clamp to floor), consistent with `find_files`.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


joryirving:
Review verdict: **changes requested before merge**.

The `list_tree` traversal itself looks good and the issue's discovery/security requirements are otherwise well covered, but there is one correctness gap in the dispatch-side response cap.

`execute_tool_request()` claims to apply `max_response_bytes` at row boundaries, but the current `cost = len(path) + len(type) + 4` is not a byte count and is not the serialized payload size. It undercounts UTF-8 filenames, JSON escaping, keys/quotes/separators, etc. The first row is also always admitted (`if kept and ...`), so a single sufficiently large entry can exceed the configured cap outright. That conflicts with #566's requirement to apply the harness's normal output-size truncation/capping.

There is a second metadata inconsistency in the same block: after byte-cap truncation, `entries` is shortened but `total` is still taken from the pre-cap executor result. The function's documented contract says `total` is the number of entries actually returned, so callers can receive e.g. one row with `total: 9`.

Please make the cap operate on the actual UTF-8 serialized row/result size while still truncating only at row boundaries, ensure no retained row can push the response beyond the configured cap, and keep `total` consistent with the returned entries (or explicitly introduce a separate available/pre-cap count if that's desired). A regression test with a multibyte/hostile filename plus a tiny cap should pin the byte semantics, and the existing byte-cap test should assert the `total` value too.

Everything else I checked looks nicely scoped to #566, and CI is green.

_Note: GitHub does not allow me to submit a formal REQUEST_CHANGES review on a PR authored by the same account, so this is posted as a PR comment instead._

joryirving:
Fixed in commit 43f0261:

1. **Byte cap now measures actual UTF-8 serialized size**: `len(json.dumps(e, separators=(",", ":")).encode("utf-8"))` — includes JSON keys/quotes/separators, not char count
2. **No unconditional first-row admission**: removed the `if kept and ...` guard so the cap is always enforced — a single large entry yields an empty result with `truncated: true`
4. **`total` consistent with returned entries**: now `len(entries)` rather than `res.get("total", len(entries))`

Added 3 regression tests:
- `test_list_tree_via_executor_byte_cap`: cap raised from 20 → 70 bytes to test two-row behavior; adds `total` assertion
- `test_list_tree_via_executor_byte_cap_multibyte_filename`: `café.txt` (= 34 UTF-8 bytes > 20-byte cap) must be excluded — this test would fail against the old char-counting logic
- `test_list_tree_via_executor_byte_cap_no_first_row_slip`: ensures a single entry exceeding the cap produces an empty result

CI is green, AI reviewer approved.

joryirving:
Follow-up review on `43f0261`: the original two issues are fixed (first-row slip is gone, `total` now matches returned entries), and the new regression coverage is useful. CI is green.

There is one remaining edge in the hard byte-cap implementation before I’m comfortable merging it: the loop sums `len(json.dumps(e, ...).encode("utf-8"))` for each individual row, but the actual returned `entries` value is a JSON array. Multiple retained rows require at least the comma separators (and, if the contract is the serialized `entries` payload itself, the surrounding `[]`). So e.g. two 34-byte rows with a 68-byte cap are both retained by the current code, while serializing those two rows as an array is >68 bytes.

Please account for the inter-row separator bytes (and ideally pin exactly what object the cap applies to, then measure that representation). A tight-boundary regression test where `sum(row_sizes) <= cap` but `serialized_array_size > cap` would catch this. Once that’s addressed, I’m good to merge this.

joryirving:
Fixed in commit b64e8be: the byte cap now measures the serialized `entries` array, not just the sum of row sizes.

The accumulator starts at 2 (the `[]` brackets) and adds `row_bytes + (1 if kept else 0)` for each row, so inter-row separators are included. Two 30-byte rows with a 60-byte cap: sum(row_sizes) = 60 <= 60, but the serialized array is [a,b] = 63 > 60 — the second row is correctly dropped.

Added `test_list_tree_via_executor_byte_cap_array_separators` as a tight-boundary regression test that would have failed against the old code.

All 1794 tests pass.

PR #599, opened 2026-09-14
feat(context): add repository map context

### LINKED ISSUE #570: [P1] Context: wire repository map into the review corpus and native-loop opening context

joryirving:
Closes #570.

Generate a bounded repository map for each review, surface it early in the corpus, and provide the native tool loop an explicitly framed opening-context slot before its first tool decision. Configuration changes are fingerprinted; disabled or failed generation clears artifacts and continues without a placeholder.

Verified with python3 -m pytest tests/ -q.


joryirving (commented):
Blocking feedback on `81e4f995` before merge. The overall wiring is good, but there are two production-path budget bugs that the current tests miss.

### 1. The repository map can crowd mandatory planning context out of the first native-loop turn

`build_planning_context()` still reserves only `_PLANNING_RESERVE = 2200` bytes for the diff head, and the new plan order is:

`Classification -> Repository Map -> Changed Files -> Version Hints -> Standards`

This repo dogfoods with `tool_corpus_max_bytes: "15000"`. That leaves only ~12.8 KB for all pre-diff sections. A valid near-cap repository map can consume ~12 KB immediately after classification; the next iteration then hits `avail < 400` and breaks before Changed Files, Version Hints, and, most importantly, Repository Standards are admitted.

That violates #570's explicit requirement that the map is an orientation aid and **must not crowd out preserved repository standards**. It also regresses the planner's existing high-signal context contract, which says it needs changed files and standards to decide what evidence to gather.

Please make the map budget adaptive to the shared planning budget so downstream mandatory sections retain space (or otherwise reserve/reorder those sections). Add a regression using the real dogfood-sized planning budget (`15000`) plus a ~12 KB map and assert that standards still reach the first-turn context; I would also pin Changed Files while you're there.

### 2. `REPO_MAP_MAX_BYTES` is bypassed when the map comes from the corpus, which is the production path

The new per-map cap is only enforced by `_repo_map_excerpt()`. But production calls:

```py
build_planning_context(corpus_max_bytes, Path("review-corpus.truncated.md"))
```

and the corpus-region fast path does:

```py
region = regions.get(region_key)
if region is not None and len(region.encode("utf-8")) + 2 <= avail:
    section = region
```

That check only compares the map region to the *shared* remaining budget; it never compares it to the plan entry's `cap` / `REPO_MAP_MAX_BYTES`. So a corpus `# Repository Map` region larger than the configured map budget is embedded verbatim whenever it fits the overall planning budget.

The existing `test_repo_map_is_bounded_by_configured_bytes` and tiny-budget test do not pass a corpus path, so they only exercise the fallback `_repo_map_excerpt()` branch and cannot catch this.

Please require the corpus map region to fit `min(REPO_MAP_MAX_BYTES, avail)` or fall back to the bounded map excerpt, and add a corpus-path regression that sets a small map cap and provides a larger `# Repository Map` region.

Related: `corpus.sh` caps `repo-map.md` first and then adds a replacement header + trust-boundary line around it, so the final model-facing corpus section can exceed `REPO_MAP_MAX_BYTES` too. The new planner tests treat the configured value as a cap on the **framed model-facing map section**, so the corpus path should follow the same contract.

Everything else I checked looks solid: generation is advisory and stale-safe, fingerprinting is wired, artifacts are guarded, corpus placement is early, and CI is green.

joryirving:
@joryirving Fixed the two blocking budget issues in b33c927.

- Native-loop mandatory planning sections now receive their bounded slots before the repository map; the map uses only remaining budget and is inserted after classification. A 15KB-budget regression retains Changed Files and repository standards with a near-cap map.
- Corpus-region embedding now enforces `min(REPO_MAP_MAX_BYTES, remaining_budget)` and falls back to the bounded artifact if needed. The corpus also caps the complete framed map section, not only the source artifact.

Verified: `python3 -m pytest tests/ -q` (1912 passed, 2 subtests) and `bash tests/test_artifact_paths_drift.sh`.

joryirving:
One remaining blocker before I’m comfortable merging `b33c927`:

### The framed corpus cap can break the repository map’s Markdown fence

The two original budget issues are fixed. However, `build_bounded_repo_map()` now re-truncates the already fence-safe `repo-map.md` by slicing bytes to a budget, backing up to the last newline, and appending `[truncated]`.

That preserves the byte cap, but it does **not** preserve the tree-fence invariant that `render_repo_map_markdown()` deliberately guarantees.

Concrete case: take a valid source map that is itself under `REPO_MAP_MAX_BYTES`, but only exceeds that same cap after the 89-byte corpus trust prefix is added. If the amount that must be shaved lands inside the ````text tree block, this helper drops the closing ```` line and appends `[truncated]` inside the still-open fence. Following corpus sections can then be interpreted as part of the map’s code block.

I reproduced this with a synthetic valid map: source length 1446 bytes, cap 1470 bytes. The source is under-cap; after framing it needs trimming. Equivalent `build_bounded_repo_map()` logic returned 1464 bytes with one ````text opener and **zero** bare ```` closing lines.

Please preserve the renderer’s fence-safe truncation contract when applying the framed cap. Good options:

- account for the trust-prefix bytes before rendering so `render_repo_map_markdown()` gets the actual body budget; or
- extract/reuse the existing fence-aware truncation logic for the framed document rather than generically re-slicing rendered Markdown.

Please add a regression where framing overhead forces the second-stage cap to land inside the tree and assert both the hard byte cap and a closed fence.

Everything else on `b33c927` looks good: the map is allocated only after the essential planning sections, production corpus-region embedding now honors `REPO_MAP_MAX_BYTES`, stale artifacts remain cleared, the new artifact path is guarded, and CI + the incremental AI review are green.

joryirving:
@joryirving Pushed the fence-safe fix as unsigned commit e122ec0.

The renderer now receives a body budget net of the trust-framing overhead, and corpus construction only re-frames its output rather than slicing Markdown a second time. Native planning takes the same fence-safe re-rendering path. Regression coverage forces a real tree-fence cut after framing and asserts both the final hard cap and balanced four-backtick fences.

Also corrected the tiny-cap boundary so a cap that cannot carry the trust prefix plus the renderer’s smallest safe result skips the advisory map.

Verified: `python3 -m pytest tests/ -q` (1916 passed, 2 subtests), `bash tests/test_context_budget.sh`, and `bash tests/test_corpus_standards_survival.sh`. `tests/test_artifact_paths_drift.sh` does not run on this macOS checkout because its GNU-sed-only `\+` extraction expression returns no guard paths; no artifact-path changes were part of this commit.

Issue #572, opened 2026-09-12
[P1] Context: build bounded related-code context from change anchors


joryirving:
## Depends on

- #571 — deterministic `change-anchors.json`

This issue should consume #571's merged artifact contract rather than invent a second diff parser.

## Summary

Build a deterministic related-code scanner that turns high-signal change anchors into a bounded repository relationship summary: likely references/callers, matching tests, and nearest package/config manifests.

This issue produces `related-code.json` + `related-code.md`. It does **not** wire them into the review corpus/native loop yet.

## Goal

Before the LLM starts its own tool exploration, we want a cheap answer to questions such as:

- Where else is this changed function/class referenced?
- Which tests appear to cover this file or symbol?
- Which package/project manifest governs this changed file?
- Is this helper used by more than the changed file suggests?

This should be a **relationship overview**, not an attempt to inline the full contents of every related file.

## Inputs

Consume:

- `change-anchors.json` from #571;
- current checked-out Git worktree;
- optionally `pr-files.json` to identify changed paths and avoid reporting the diff itself as a "related" reference.

No model or network calls.

## Output

Versioned JSON plus compact Markdown.

Illustrative JSON shape:

```json
{
  "version": 1,
  "files": [
    {
      "path": "pr_reviewer/tool_executors.py",
      "symbols": [
        {
          "name": "git_grep",
          "references": [
            {"path": "scripts/run_tool_harness.py", "line": 412, "snippet": "..."}
          ]
        }
      ],
      "tests": ["tests/test_gh_api.py", "tests/test_run_native_loop_wiring.py"],
      "manifests": ["requirements.txt"]
    }
  ],
  "truncated": false
}
```

The schema may differ, but version it and preserve attribution back to the changed file/anchor.

## Reference lookup

For high-confidence symbol anchors, use safe argv-only Git searches.

Prefer fixed-string matching unless there is a strong reason not to:

```text
git grep -n -F -- SYMBOL -- .
```

Requirements:

- bound hits per symbol (suggest ~20);
- bound total symbols searched (suggest ~40 high-confidence symbols globally);
- bound total reference rows globally;
- omit obvious self-reference noise where the only hit is the defining changed line;
- retain `path:line` and a short bounded snippet;
- stable ordering by anchor order/path/line.

Do not attempt semantic call-graph correctness. Label these as "references" rather than claiming every textual hit is a runtime caller.

## Test discovery

Use deterministic path/name heuristics first, then symbol references inside likely test directories.

[...]

PR #600, opened 2026-09-15
feat(context): add related code scanner

joryirving:
Implements #572 with a bounded, deterministic related-code artifact seeded from change anchors. It scans high-confidence symbols for textual references, likely tests, and nearest manifests without executing repository code or making network/model calls.


joryirving:
I would **not merge this head yet**, but my blockers differ from most of the automated path findings.

I do **not** consider the explicit CLI `--workspace` / `--anchors` / `--files` behavior a path-traversal vulnerability here. Those are caller-selected CLI inputs, not PR-controlled paths, and `test_cli_writes_versioned_json_and_markdown` intentionally passes an anchor file outside the scanned repository. Likewise, `_resolve_output()` resolves both the workspace root and target before containment, so a symlinked workspace root does not by itself escape the resolved workspace. I would not contort the CLI around those findings unless the intended contract changes to “all inputs must live inside GITHUB_WORKSPACE.”

I *do* see two blocking correctness/resource issues on `ed061bd`:

1. **The per-symbol grep cap is applied before changed/self hits are filtered, so a noisy changed file can hide every real external reference.** `git_grep_references()` stops/kills Git after `max_references_per_symbol + 1` raw rows. Only afterwards does `build_related_context()` remove `hit["path"] in changed_paths`. Example: changed `a.py` contains `target` 21+ times and unchanged `z.py` references `target` once. Git emits the `a.py` rows first, the scanner kills grep at 21, then filters all 21 away and reports no external references. That defeats the core purpose of #572 and also prevents test discovery from seeing later test-file references. Please apply the cap to *eligible related rows* (e.g. pass excluded changed paths into the streaming grep and continue until N+1 non-excluded hits), with a real-Git regression for this case. The current `omitted_references` accounting is also misleading for the same reason: it increments based on raw rows that may all be self/change noise.

2. **The artifact is not actually hard-bounded because manifest ownership is unbounded, and JSON has no output cap.** `_discover_manifests()` returns every matching manifest at every ancestor with no per-file/global cap. A PR can add thousands of root `requirements-*.txt` / `Dockerfile*` files; each changed file then repeats that entire list in `related-code.json`. Markdown happens to have `MAX_MARKDOWN_BYTES`, but the versioned JSON artifact can balloon arbitrarily, contrary to #572’s “caps on ... output bytes” / “hard-bounded and explicit when truncated” acceptance criterion. Add a deterministic manifest cap (and explicit omitted accounting) and make the JSON/output budget structurally hard-bounded rather than relying on the Markdown renderer alone.

Everything else I checked is directionally good: fixed-string argv-only grep, high-confidence symbol cap, snippets are redacted/bounded, tests are capped, manifest ordering is nearest-first, hostile Markdown is code-span escaped, and CI is green.

joryirving:
Pushed commit 977f465 to address the valid review feedback.

- Reference streaming now excludes changed paths before applying its cap; a real-Git regression proves noisy changed files no longer hide external/test references.
- Added an explicit per-file manifest cap plus a structural, valid-JSON 100 KB hard cap and truncation accounting.
- JSON load failures now emit generic errors rather than raw OS error details.

I intentionally did not change explicit CLI workspace/input containment or symlink behavior: those are caller-selected paths, and the existing CLI contract intentionally supports anchors outside the scanned repository.

Verified: python3 -m pytest tests/ -v --tb=short — 1922 passed, 2 subtests passed.

joryirving:
One remaining blocker on `977f465` before I’m comfortable merging:

### Global reference cap can silently skip later symbols without marking truncation

The fix removed the old `remaining <= 0` truncation branch. Now, once `references_total == max_references`, later symbols simply skip `git_grep_references()`:

```py
remaining = max_references - references_total
...
if remaining > 0 and max_references_per_symbol > 0 and not tracked_error:
    hits, additional_hit, grep_error = git_grep_references(...)
```

If the earlier symbol(s) consumed the global cap **exactly** without an `additional_hit`, the artifact can remain `truncated: false` even though every later high-confidence symbol is emitted with `references: []` without ever being searched.

Concrete case: `max_references=1`, symbol A has exactly one external reference, symbol B also has external references. A consumes the one global slot with no overflow signal; B is never searched; no `reference_cap` reason is recorded. That makes “no references” indistinguishable from “not searched because the global budget was exhausted,” which violates #572’s requirement that bounded behavior be explicit.

Please restore an explicit global-cap signal when there are remaining selected symbols that cannot be searched due `remaining <= 0`, and add a regression for the exact-cap-then-later-symbol case. The omitted count does not need to pretend it knows the exact unseen reference count, but the artifact must at least mark `reference_cap` / truncated rather than silently claiming completeness.

Everything else from my previous review looks fixed: changed-file hits are excluded during streaming before the eligible cap, manifests are capped, JSON is structurally hard-bounded, and exact-head CI is green.

joryirving:
Pushed the global-cap fix in commit `50dc981`.

When an earlier symbol consumes the global reference budget exactly, later symbols now mark the artifact truncated with `reference_cap` rather than silently appearing to have no references. The omitted count records the unsearched symbol as an explicit unknown. Added the exact-cap-then-later-symbol regression.

Verified: `python3 -m pytest tests/ -v --tb=short` — 1923 passed, 2 subtests passed.

PR #603, opened 2026-09-15
test(evals): add repository context fixtures

### LINKED ISSUE #577: [P1] Evals: add repository-context capability fixtures and scoring
## Depends on

Core fixtures should target merged capabilities from:

- #566 `list_tree`
- #567 `find_files`
- #570 repository-map integration
- #573 related-code integration

The eval framework work can be prepared earlier, but do not mark the new capability checks complete until those features exist.

## Summary

Extend the existing agentic eval corpus/scorer with repository-context scenarios that measure whether the reviewer actually uses structural/relationship context to improve review quality.

The goal is to prevent the new context work from becoming prompt decoration that never changes model behavior.

## Existing foundation

The repo already has `evals/corpus-agentic.json` and `scripts/eval_harness.py` with capability checks for cases such as tool-required verification, dependency support-matrix lookup, trivial digest updates, and hardcoded-secret detection.

Reuse that harness and its capability-altitude philosophy: the **grader** may know the expected evidence; the reviewer prompt must not name the exact tool recipe.

## New fixture categories

Add at least four repository-context scenarios.

### 1. Tree discovery required

Synthetic or stable fixture where the diff references a subsystem/file indirectly and the reviewer cannot know the relevant path from the diff alone.

Expected evidence should require either:

- `list_tree`, or
- `find_files`,

followed by a read of the discovered file.

The reviewer prompt must not tell it the path.

### 2. Related test discovery

Fixture where a changed implementation has a non-obvious regression covered by an existing test outside the changed file.

Expected evidence should verify that the reviewer consulted or explicitly reasoned from the `Related Code Context` test candidate and, when needed, read the test file.

### 3. Cross-file reference/caller impact

Fixture where a helper/API shape change breaks an existing reference in another source file that is not part of the PR diff.

Expected finding: the reviewer identifies the downstream reference/contract issue. The grader may require mention/read of the related file.

### 4. Repo-map orientation without unnecessary tool spam

Fixture where the repository map is enough to establish where tests/config live and the diff itself is correct.

Expected behavior: approve/clean review without wasting the entire tool budget. This is the negative control preventing "more context" from turning into mandatory browsing.

## Optional fifth fixture: related repository

After #576 lands, add a separate scenario where an explicitly allowlisted sibling repo contains the interface contract needed to review the PR. Keep this optional so the local full-repo-context milestone is not blocked on remote-repo work.

## Scoring additions

Extend the eval result schema only as needed to capture:

- expected tool names;
- expected discovered/read paths;
- expected review mentions/finding text;
- maximum acceptable unnecessary tool calls for the negative-control fixture.

Prefer generic checks rather than model-specific prose matching.

If the current evaluator can already express these with `tool_call`, `args_contains`, and `review_mentions`, add no new scoring type unless necessary.

## A/B modes

For the new corpus, compare at least:

- `tools_off` baseline;
- `native_loop` with the new repo context enabled;
- where practical, native_loop with repo-map/related-code context disabled to isolate the deterministic context value from the new tools themselves.

If adding a new harness mode would be too invasive, expose the context toggles through the eval workflow/matrix instead.

## Model / run count

Keep the existing guidance that one local-model run is noisy. Use repeated runs when measuring capability pass rates.

The primary target model for dogfood is Qwen3.8-27B-class, but fixtures must remain model-agnostic and useful for future models.

## Expected files

Likely:

- `evals/corpus-agentic.json` or a new `evals/corpus-repo-context.json`
- `scripts/eval_harness.py`
- `.github/workflows/eval-harness.yaml`
- tests for evaluator/scorer changes
- `AGENTS.md` / README eval documentation if the new corpus/matrix is user-visible

Prefer a separate corpus file if that keeps the existing benchmark stable and allows repo-context work to evolve independently.

## Required tests

- fixture schema validation;
- each new expected-evidence check can both pass and fail on synthetic traces;
- path/tool checks are not satisfied by unrelated calls;
- negative-control unnecessary-tool threshold is deterministic;
- old agentic corpus still scores identically after evaluator changes;
- workflow can select the new corpus/config matrix.

## Acceptance criteria

- [ ] At least four repo-context scenarios exist with concrete expected evidence.
- [ ] Reviewer prompts stay at capability altitude and do not reveal the answer/path/tool sequence.
- [ ] Scorer can distinguish successful repository discovery from a diff-only rubber stamp.
- [ ] At least one negative control penalizes unnecessary exploration.
- [ ] Existing eval corpus behavior is unchanged.
- [ ] Eval workflow can run the new scenarios repeatedly against the configured model.

## Non-goals

- Establishing a universal benchmark score target in this PR.
- Changing production review defaults based on a single run.
- Building a new evaluator framework from scratch.

joryirving:
Closes #577.

Adds a separate repository-context corpus for the merged context and tooling PRs, plus deterministic scoring for a tool-call budget. The existing workflow can select the new corpus without changing its scheduled baseline.

Verified: full pytest suite passed (1959 tests and 2 subtests).


joryirving:
Blocking issue: the three positive repo-context fixtures currently grade files that are already present in each historical PR's changed-file list, so they do not actually measure the repository-context capability #577 asks for.

Concrete examples:
- PR #598 changes `pr_reviewer/tool_executors.py`, but the tree-discovery fixture grades a `read_file` of `tool_executors`. The reviewer already gets that path from the PR diff, so `list_tree`/`find_files` is not needed to discover it.
- PR #600 changes `tests/test_related_context.py`, but the related-test fixture grades reading that same changed test. #577 specifically calls for a non-obvious existing test outside the changed file/diff.
- PR #601 changes `scripts/sections/corpus.sh`, but the cross-file caller fixture grades reading `corpus.sh`. #577 specifically requires an existing downstream reference in another source file that is not part of the PR diff.

This means the corpus can pass while the reviewer is still effectively diff-local, which defeats the core acceptance criterion: distinguish successful repository discovery from a diff-only rubber stamp.

Please replace/rework the three positive fixtures so the required discovered/read path is genuinely outside the selected PR's changed files. Add a regression that asserts each positive fixture's graded target path is not in that fixture PR's changed-file set (hardcoded historical sets are fine if we want deterministic/offline tests). Also make the discovery check itself path-relevant enough that an unrelated `list_tree`/`find_files` call plus a lucky read cannot satisfy the scenario.

The `max_tool_calls` scorer and negative-control fixture look good.

PR #604, opened 2026-09-15
feat(tools): add repo contents reader

### LINKED ISSUE #576: [P2] Tooling: add an allowlist-gated repo_contents tool for related repositories
## Summary

Add a native-loop `repo_contents` tool that can list a directory or read a text file from an explicitly allowlisted related repository.

The action already has `tool_allowed_gh_api_repos`, but using raw `gh_api` contents endpoints is awkward for models because file responses may be encoded and directory/file shapes differ. This tool should provide one small, normalized related-repo primitive.

## Desired contract

Suggested arguments:

```json
{
  "repo": "owner/repository",
  "path": "src/client",
  "ref": "main"
}
```

- `repo` required.
- `path` optional; empty means repository root.
- `ref` optional; default should be the target repository's default branch/API default rather than guessing `main`.

Behavior:

- directory path -> return a bounded sorted listing with names/paths/types;
- text file path -> return decoded text with the same general size cap as local `read_file`;
- binary/non-text content -> return metadata/clean error, not raw binary;
- do not recursively download entire repositories.

## Authorization boundary

Reuse the existing `gh_api` repository allowlist/security logic. The tool may access only:

- the current repository; or
- repositories explicitly present in `tool_allowed_gh_api_repos`; or
- `*` only when the operator has deliberately configured the existing wildcard behavior.

Do not create a second allowlist implementation.

The model-provided `repo` must never become a host name or arbitrary URL. Requests go only through the active platform backend / existing validated API path.

## Platform behavior

Build this on the existing platform-aware `gh_api` seam where practical.

GitHub and Forgejo may return slightly different contents payloads; normalize them into one tool result shape. If the current Forgejo backend cannot support an operation, fail with a clear unsupported result rather than silently falling back to github.com.

Do not add a broad new Forgejo API client in this issue.

## Result shapes

Directory example:

```json
{
  "repo": "owner/repository",
  "path": "src",
  "type": "directory",
  "entries": [
    {"path": "src/api", "type": "dir"},
    {"path": "src/client.py", "type": "file"}
  ],
  "truncated": false
}
```

File example:

```json
{
  "repo": "owner/repository",
  "path": "src/client.py",
  "type": "file",
  "content": "...bounded text...",
  "truncated": false
}
```

Exact schema may differ, but keep it stable and easy for a 27B model to consume.

## Bounds

Suggested limits:

- directory entries: 200, clamp max ~500;
- decoded file content: reuse existing tool result/output cap (~12 KB);
- API timeout: reuse `tool_request_timeout_sec`.

Stable sorting is required.

## Tool schema guidance

The model-facing description should explain when to use this tool:

- inspect an explicitly related API/client/shared-library repo;
- verify a cross-repo interface/manifest/version contract;
- do not browse unrelated repositories speculatively.

The planning context should continue to expose which repositories are allowed so the model does not waste calls probing denied repos.

## Expected files

Likely:

- `pr_reviewer/conversation.py`
- `pr_reviewer/platform.py` only if a small normalized contents helper belongs there
- `pr_reviewer/forgejo_backend.py` only if current platform translation needs a focused contents mapping
- `scripts/run_tool_harness.py`
- `README.md`
- `AGENTS.md`

Tests:

- new/extended tool executor tests
- `tests/test_run_native_loop_wiring.py`
- existing platform `gh_api` GitHub/Forgejo tests

## Required tests

- current repo allowed;
- explicitly allowlisted second repo allowed;
- unlisted repo rejected;
- wildcard follows existing wildcard semantics;
- directory payload normalized/sorted/capped;
- text file decoded and byte-capped;
- binary/unsupported content handled cleanly;
- missing path/404 handled cleanly;
- API timeout handled;
- model schema/dispatch round trip;
- Forgejo path either works through the platform seam or returns explicit unsupported behavior with no github.com fallback.

## Acceptance criteria

- [ ] Native-loop reviewer can list/read an explicitly permitted related repo through one normalized tool.
- [ ] Existing repo allowlist remains the single authorization source of truth.
- [ ] Tool cannot choose an arbitrary host/URL.
- [ ] File/directory results are bounded and deterministic.
- [ ] GitHub/Forgejo platform behavior is explicit and tested.
- [ ] Existing `gh_api` functionality remains unchanged.

## Non-goals

- Organization-wide repository discovery.
- Remote semantic/code search.
- Cloning related repositories.
- Automatically choosing which repos are related.
- Write operations.

joryirving:
Adds an allowlist-gated `repo_contents` native tool so reviews can inspect a related GitHub repository without arbitrary host or URL access. It reuses `tool_allowed_gh_api_repos`, returns bounded deterministic file/directory results, and fails explicitly on Forgejo rather than falling back to GitHub.

Closes #576

Tests: `python3 -m pytest tests/test_conversation.py tests/test_platform_gh_api_forgejo.py tests/test_tool_executors.py tests/test_tool_request_normalization.py tests/test_run_native_loop_wiring.py -q -k "not test_native_loop_forwards_git_grep_optional_args"` (233 passed; unrelated deselected test cannot create its fixture commit because the local 1Password SSH signing agent errors).


joryirving:
Blocking security issue in `repo_contents`:

`_validate_repo_contents()` blocks sensitive *requested path names*, but GitHub's Contents API dereferences an in-repo symlink when the symlink target is a normal file. That means an allowed repo can contain e.g. `docs/config -> ../.env`; a model request for `docs/config` passes `SENSITIVE_PATH_RE`, and GitHub returns the target file as a normal `type: file` response with its content.

GitHub documents this behavior explicitly: for a symlink whose target is a normal file in the repository, the contents endpoint responds with the target file's content.

Please make the remote sensitive-file boundary robust to symlink aliases rather than relying only on the model-supplied path. A good approach is to verify the returned object's canonical/target path or use an API shape that lets us identify symlink resolution before exposing content, then re-run the sensitive-path check against the resolved target. If that cannot be determined safely, fail closed for symlink-resolved file responses.

Add a regression where a safe-looking path aliases a blocked target (e.g. `.env`) and assert no content is returned.

The current automated review's statement that the GitHub API does not resolve symlink targets is incorrect; GitHub's REST docs say the opposite.

joryirving:
Blocking security issue in `repo_contents`: the direct file-read path can bypass `SENSITIVE_PATH_RE` through a repository symlink.

`_validate_repo_contents()` only checks the model-supplied path. `_repo_contents_github()` then calls GitHub's Contents API directly for that path and trusts a returned `type: "file"` payload. GitHub explicitly documents that when a requested path is a symlink whose target is a normal file in the repository, the Contents API responds with the *target file's content*, not a symlink object:
https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content

So, for example, a permitted repo could contain `docs/config-link -> ../.env`. `repo_contents(repo, "docs/config-link", ...)` passes the sensitive-path check because the requested path is innocuous, but GitHub can return `.env` contents as a normal file response. That makes the new sensitive-path boundary bypassable.

Please make direct file reads fail closed on symlinks before returning content. A straightforward implementation is to inspect the requested entry via its parent directory listing (or another non-following Git object/tree primitive), reject/return metadata for `symlink` and `submodule`, and only fetch content once the entry is proven to be an ordinary file. Do not rely on the direct Contents response's `type` to distinguish this case.

Add a regression where the requested safe-looking path is a symlink to a blocked sensitive target and assert that target content is never returned. Also add a normal symlink case so the tool's direct-symlink behavior is explicit.

The automated review's statement that GitHub's Contents API does not resolve symlink targets is incorrect; the official API docs say the opposite.

joryirving:
One security blocker remains in the symlink preflight design: the parent-listing preflight and the subsequent direct content fetch are two separate requests against the same *ref string*, not the same immutable revision.

For a branch/default ref, this leaves a TOCTOU window:

1. Parent listing at commit A reports `docs/config-link` as `type: file`, so preflight passes.
2. The branch moves to commit B where `docs/config-link` is a symlink to `.env`.
3. The direct Contents API request runs against commit B. GitHub documents that when a symlink targets a normal file in the repo, the Contents API returns the target file's contents, so the sensitive-path boundary can still be bypassed.

`test_repo_contents_regular_file_preflight_uses_same_ref` only proves both requests carry the same symbolic ref (`release`); it does not prove they observe the same commit.

Please pin object identity before returning file content. Two clean options:

- Resolve the supplied/default ref once to an immutable commit SHA and use that SHA for both parent preflight and direct content fetch; or
- Preferably for files, use the parent listing's ordinary-file `sha` and fetch that exact blob through the Git blob endpoint instead of doing a second path-based Contents API read. That removes both the symlink-follow behavior and the race.

Add a regression that simulates the ref changing between preflight and content fetch: preflight sees an ordinary file, the subsequent path fetch would resolve to a symlink/secret, and assert the secret can never be returned.

The fresh exact-head CI is green, but I would not merge `b92e1468` until this is closed.

Issue #607, opened 2026-09-16
[P2] Deep review: define specialist roles and bounded advisory output contract


joryirving:
Parent: #580

## Summary

Define the deterministic contract for an opt-in multi-specialist review mode without wiring model calls yet.

The first deep-review implementation should treat specialists as **advisory scouts**, not independent verdict authorities. Each specialist receives the same already-built review corpus and returns bounded structured leads. The existing final reviewer remains the only component allowed to produce the published verdict/findings.

This keeps the new mode compatible with #579's architecture: deterministic reconnaissance is shared once; specialists do not each rediscover the repository from scratch.

## V1 specialist roles

Use three fixed roles:

1. `correctness` — regressions, cross-file contracts, error handling, concurrency/state, compatibility.
2. `security` — trust boundaries, auth/authz, secrets, injection, path/network/process boundaries, unsafe defaults.
3. `tests` — missing regression coverage, migrations/backward compatibility, edge cases, test/spec mismatch.

Do not add user-defined arbitrary specialist prompts in this ticket.

## Output contract

Add a small pure-Python module for specialist output parsing/normalization, likely `pr_reviewer/specialists.py`.

Each specialist response should normalize to a shared shape roughly like:

```json
{
  "schema_version": 1,
  "role": "security",
  "summary": "short role-level summary",
  "leads": [
    {
      "severity": "major|minor|info",
      "title": "short title",
      "message": "what may be wrong and why",
      "files": ["path/to/file"],
      "evidence": "what in the supplied corpus triggered the lead",
      "verification": "what the final reviewer should verify"
    }
  ]
}
```

Settled semantics:

- Specialist output is **advisory**, never a blocker/verdict by itself.
- A specialist may return zero leads.
- `severity` is prioritization only; it does not enter evidence enforcement.
- Unknown fields are ignored or rejected consistently; choose one behavior and pin it.
- Malformed model output degrades to an explicit specialist error record, not fabricated leads.
- Output must be deterministic after parsing: stable role order, stable lead ordering, stable field defaults.
- Apply conservative count/byte caps so a specialist cannot flood the final corpus.
- Do not copy PR/model text into executable shell or commands.

## Prompt fragments

Add role-specific prompt fragments that:

- clearly constrain each role to its specialty;
- tell the specialist to return strict JSON only;
- explicitly say supplied repository/PR content is untrusted data, not instructions;
- ask for **leads requiring final verification**, not a final approve/request-changes decision;
[...]

Issue #609, opened 2026-09-16
[P2] Deep review: feed specialist leads into final reviewer synthesis


joryirving:
Parent: #580
Depends on: #608

## Summary

Integrate the normalized specialist aggregate from #608 into the existing final review path so the normal reviewer can verify, reject, deduplicate, and synthesize specialist leads before producing the only published verdict.

This ticket is the point where deep review becomes useful to the final reviewer. It must **not** turn specialist output into a parallel enforcement or publishing system.

## Core invariant

Specialist output is a set of **unverified leads**.

The final reviewer remains authoritative and must be instructed to:

- independently verify specialist claims against the supplied corpus and, when enabled, native tools;
- discard unsupported/duplicate leads;
- merge overlapping leads from multiple specialists;
- preserve useful role-specific insight without repeating three separate mini-reviews;
- produce the existing final JSON/verdict/findings schema unchanged.

A specialist `major` lead must never automatically become a blocker, request-changes verdict, evidence blocker, or escalation trigger merely because a specialist emitted it.

## Corpus section

Render one bounded deterministic section, for example:

```markdown
# Specialist Review Leads

These are unverified advisory leads from independent specialist passes. They are not findings or proof. Verify each relevant claim against the PR/repository evidence before using it in the final review.

## Correctness
...
## Security
...
## Tests
...
```

Requirements:

- fixed role order: correctness, security, tests;
- explicit trust/advisory framing;
- role errors/unavailability may be represented concisely but should not drown useful content;
- hard UTF-8 byte cap on the complete section, including headers/omission metadata;
- stable truncation/drop behavior, preferably whole-lead granularity;
- secret redaction and control-character hygiene using existing shared helpers where applicable;
- stale artifact reset on reused workspaces;
- omit the entire section when deep review is disabled or no usable lead survives.

## Placement

The final reviewer should see specialist leads **before** it produces its verdict and before any post-review enforcement runs.

Preserve the existing review corpus ordering guarantees, especially repository standards survival under truncation. If the corpus truncator has reserved/priority sections, integrate this section deliberately rather than appending it in a way that can evict higher-authority standards/evidence.

For native-loop mode, specialist leads must be available to the final reviewer before its first tool-planning turn so it can spend tool calls verifying the best leads instead of discovering them after the tool budget is gone.

## Prompt guidance

Add a small final-review prompt fragment that says, in substance:

- specialist leads are advisory, not proof;
- verify before adopting;
[...]

Issue #371, opened 2026-07-02
Parallelize independent context-gathering phases and batch enrichment API calls


joryirving:
From the simplify sweep behind #366 (continuation of the #258 inventory; complementary to the #359 module split).

**Smell (1):** `run_review.sh` runs four independent context-gathering phases strictly sequentially: linked-source enrichment (up to 60s budget), image-digest provenance (up to 60s), evidence providers (up to 30s/provider), and the repo-impact/history git scans. None consumes another's output — all read only what context.sh already wrote, and all are consumed together at `build_review_corpus`. Worst case serializes ~2+ minutes of wall clock into every review. Fix: background jobs (`… &`; `wait`) before the corpus step.

**Smell (2):** inside `run_enrichment.py`, all `gh_api_call` invocations in Phases 2–4 run sequentially, one `gh` subprocess per call (~0.2s spawn + RTT each; 2 calls per release URL, 1–2 per repo candidate, up to 3 per ghcr image). On a link-heavy Renovate PR this exceeds the 60s `BudgetTracker`, and the budget check then **silently drops the remaining sources** — serialization costs completeness, not just latency. Fix: fan out with the `ThreadPoolExecutor` already used for Phase-1 raw fetches, and/or replace the `gh` subprocess with direct `urllib` calls with an Authorization header.

**Smell (3):** the `classification.sh` repo-impact loop runs up to 14 terms × (`git grep` over the worktree + `git log --grep -n 10` over full history) = up to 28 sequential scans for what the code itself calls the lowest-value corpus sections. Fix: one combined `git grep -E 'a|b|…'` pass and one bounded `git log` filtered per-term (or at minimum parallelize).

Note: #366 already dedupes the double releases fetch (Phase 2 vs Phase 3) via a per-repo cache.

itsmiso-ai:
Great analysis! The three smells are clearly identified with specific fixes. This is ready to be claimed by a local worker.

PR #623, opened 2026-09-18
feat(review): add deep-review specialist advisory passes (#608)

### LINKED ISSUE #608: [P2] Deep review: run bounded parallel specialist advisory passes
Parent: #580
Depends on: #607

## Summary

Add the execution layer for the three specialist roles defined by #607.

Deep review is opt-in. When enabled, the action should run the fixed `correctness`, `security`, and `tests` specialist passes against the **same already-built review corpus** before the final reviewer runs.

Specialists are advisory only: their outputs are parsed through #607's contract and written to bounded artifacts. They do not publish comments, emit final findings, or decide the verdict.

## Settled V1 execution model

Add an input such as:

```yaml
deep_review:
  default: "false"
```

Semantics:

- `false`: byte-for-byte preserve the normal review path as far as practical; no specialist calls or artifacts with substantive content.
- `true`: run exactly three fixed specialist passes: correctness, security, tests.
- Run the three passes concurrently so wall clock is close to one model call rather than three serial calls.
- Reuse the resolved **primary model/provider configuration** for all three specialist calls in v1. Do not add a second credential/model matrix in this ticket.
- Specialists receive the same bounded corpus that the primary reviewer would receive, plus only their small role prompt.
- Specialists do **not** get independent native tool loops in v1. Repository reconnaissance from #579 is already present in the shared corpus; the final reviewer retains the tool loop for targeted verification.
- Each specialist has its own request/response artifact and timeout boundary.
- A timeout/model failure/malformed response for one specialist is fail-soft: record that role as unavailable and continue with the other roles/final review.
- Deep review must remain bounded by a separate aggregate wall-clock timeout so a wedged provider cannot stall the workflow indefinitely.

## Model-call reuse

Reuse the existing model-call request/retry/timeout machinery rather than implementing a second HTTP client.

The specialist response format differs from the final verdict schema, so factor only the transport/retry pieces that are actually shareable. Do not weaken the existing final-review response contract to accommodate specialists.

## Artifacts

Use explicit guarded artifacts, likely along the lines of:

```text
specialist-correctness.request.json
specialist-correctness.response.json
specialist-correctness.json
specialist-security.request.json
specialist-security.response.json
specialist-security.json
specialist-tests.request.json
specialist-tests.response.json
specialist-tests.json
specialists.json
```

`specialists.json` should be the deterministic normalized aggregate in fixed role order.

All generated paths must participate in the repo's existing artifact symlink/write guard.

## Configuration / fingerprinting

- Add `deep_review` to config normalization and the broad/config fingerprint wherever review-affecting inputs are fingerprinted.
- Validate boolean values using existing conventions.
- Do not enable deep review by default.
- Do not change public native-loop budgets/defaults.

## Telemetry

Expose enough non-secret telemetry for later evals/debugging, at minimum:

- enabled/disabled;
- per-role success/error;
- per-role elapsed time if the current timing framework supports it;
- aggregate specialist elapsed time;
- model identifier already safe to expose through existing analysis-engine conventions.

Do not log API keys, raw auth headers, or full specialist responses solely for telemetry.

## Expected files

Likely:

- `action.yml`
- `scripts/sections/config.sh`
- `scripts/sections/review.sh` and/or a small new specialist orchestration helper
- existing model-call helper(s), only if a clean transport primitive must be factored
- `scripts/artifact_paths.sh`
- precheck/config fingerprint code
- focused shell/Python tests
- README/AGENTS input/architecture docs

## Required tests

Cover at least:

- deep review disabled performs no specialist model calls;
- enabled dispatches exactly three roles;
- roles receive the same corpus and distinct role prompts;
- calls execute concurrently (test via controlled fake delays, not timing-flaky real network);
- one specialist failure does not cancel successful siblings or final review;
- all three failures still allow the final review to run;
- malformed response is normalized through #607 rather than crashing;
- aggregate timeout is bounded;
- artifact symlink guard covers all new write targets;
- config/fingerprint changes when `deep_review` changes;
- no new native tool loop is launched for specialists;
- existing standard-mode review tests remain unchanged/green.

## Acceptance criteria

- [ ] `deep_review=false` preserves the current normal path.
- [ ] `deep_review=true` runs correctness/security/tests specialists concurrently.
- [ ] V1 specialists reuse the primary model configuration and shared deterministic corpus.
- [ ] Specialist calls are independently bounded and fail-soft.
- [ ] Normalized aggregate output is deterministic and artifact-safe.
- [ ] Specialists cannot publish or set a verdict.
- [ ] Specialists do not run independent native tool loops.
- [ ] Existing review path remains the final authority.

## Non-goals

- Feeding specialist leads into the final reviewer; that is the next ticket.
- Specialist-specific provider credentials/models.
- Dynamic/custom roles.
- Specialist-to-specialist conversation.
- Publishing specialist sections directly to GitHub/Forgejo.
- Changing escalation policy.

joryirving:
## Why

Implements #608, the execution layer for the #607 specialist contract. `deep_review` (default `false`) opts into three fixed advisory passes — correctness / security / tests — run **concurrently** over the already-built review corpus with the primary model config, launched as a background job before the final review call so the phase overlaps the review's wall clock instead of adding three serial calls to it.

Leads land in workspace-guarded artifacts plus a deterministic aggregate `specialists.json`, and are surfaced in the step summary. Advisory only this iteration: they never touch the verdict, enforcement, findings, or the published body, and specialists never run a native tool loop.

## Notes for review

- **Disabled path is pure**: `false` → no launch, no artifacts, no summary row; the normal path is unchanged. Defense-in-depth: `run_specialists.py` also no-ops on a direct invocation with the gate off.
- **Bounded twice**: per attempt `min(ai_request_timeout_sec, deadline-now)` (floored at 0.1s — curl `--max-time 0` means unlimited), and the whole phase by `deep_review_timeout_sec` (default 600). Stragglers past the deadline are cancelled (daemon threads + per-role cancel events) and recorded as timeouts.
- **Fail-soft everywhere**: a role's transport failure (one retry; timeouts and contract failures never retried), malformed output (normalized through #607), or guard refusal degrades its entry — siblings and the final review are unaffected. Exit code is 0 whenever the aggregate was written.
- **Secret hygiene + guards**: errors masked via `mask_secrets`; the API key never enters payloads/artifacts/logs (transport's 0600 curl config only). All 11 write targets resolve under the workspace root, refuse symlinks under a shared artifact lock (closes the cancel-check-vs-write race against the reaper), and are listed in `artifact_paths.sh` for the shell-side guard.
- `deep_review` / `deep_review_timeout_sec` are added to `_EXACT_CONFIG_KEYS` so a toggle invalidates a stale review.
- Transport reuse is `pr_reviewer.transport.run_chat_request` — no second HTTP client.

## Tests

- `tests/test_run_specialists.py` + `tests/test_run_specialists_payloads.py` (34): disabled-gate purity, exactly-three-concurrent (timing-proven), identical-corpus/distinct-prompts, fail-soft siblings, timeout-not-retried, deadline reaping, retry-then-success, malformed→degraded, symlink/escape guards, secret-leak sweep, payload matrix (OpenAI/Anthropic/stream/response_format/token-field/temperature), aggregate determinism.
- `tests/test_deep_review_wiring.sh` (42 static checks): launch-before-final-reviewer, guarded reap, guard-list coverage, env-block parity, fingerprint keys.
- Full local suite: 2237 pytest + all bash suites + smoke test green. (`test_artifact_paths_drift.sh` fails locally only on BSD sed — verified identically broken on pristine HEAD; passes under CI's GNU sed.)

Closes #608


joryirving (commented):
Two blockers before merge:

1. The deep-review phase currently overlaps the final reviewer call.

#608's settled execution model says the three specialists run concurrently over the shared corpus **before the final reviewer runs**. The current review.sh launches `run_specialists.py` in the background, immediately proceeds into the final reviewer, and only reaps the specialist phase near the step summary.

That changes the phase contract and makes #609 awkward, because #609 needs `specialists.json` available before final-review synthesis. Make the three specialist calls concurrent with each other, but complete/reap the specialist phase before entering the final reviewer path. Keep fail-soft behavior: specialist failures/timeouts must not prevent the final reviewer from running.

Update the wiring/static tests/docs that currently explicitly assert or describe overlap.

2. `scripts/run_specialists.py` catches `BaseException` in the worker wrapper. That swallows `KeyboardInterrupt`, `SystemExit`, and other process-control exceptions, and CodeQL is correctly flagging it. Catch `Exception` instead.

Also make the last-resort worker-exception path obey the same artifact contract as other fail-soft outcomes: write the normalized per-role artifact (and response/error artifact as appropriate) rather than only populating the in-memory aggregate entry.

The rest of the implementation shape looks strong: exact three fixed roles, same corpus/distinct prompts, transport reuse, independent/aggregate bounds, deterministic aggregate order, workspace artifact guards, fingerprinting, disabled-path purity, and broad focused tests. Keep the current scope and avoid unrelated cleanup.

joryirving:
Both review blockers addressed in 8371f8c:

**1. Phase ordering.** The specialist phase now completes *before* the final reviewer path enters: `harvest_specialist_phase` (the #371 idiom — guarded `wait`, phase log surfaced, fail-soft) is defined and called immediately after the launch, ahead of `PRIMARY_OK`/`NATIVE_VERDICT_USED`/`call_model_tier primary`; the old trailing reap near the step summary is gone. Roles remain concurrent with each other via the runner's internal threads — exactly three fixed passes, one wall-clock call's worth of latency — but the reviewer no longer overlaps them, so #609 finds `specialists.json` fully written when final synthesis starts. `test_deep_review_wiring.sh` now pins it: launch and reap both precede the native-verdict path and the primary call, and exactly one `wait "$SPECIALISTS_PID"` exists. Comments in `review.sh`, the `run_specialists.py` docstring, AGENTS.md, and the README timeout row now describe the sequential-with-respect-to-the-reviewer model. Fail-soft unchanged: a timed-out/failed phase records errors and the final reviewer runs regardless.

**2. Worker exception guard.** `except BaseException` → `except Exception` in the thread wrapper (KeyboardInterrupt/SystemExit propagate — CodeQL finding resolved). The crash path now honors the same fail-soft artifact contract as every other failure mode: it writes `specialist-<role>.json` (empty #607 contract artifact with the masked crash error in `errors`) and `specialist-<role>.response.json` (`{"error": ...}`) before populating the aggregate entry — factored into `_write_role_failure_artifacts`, shared with the deadline reaper so both write the identical pair (cancel-aborted for workers, unconditional for the reaper). New test `test_worker_crash_writes_full_artifact_set` patches `_run_role` to raise and asserts both files, the entry, and the aggregate.

Re-verified after the change: 2238 pytest (35 on the two specialist suites, twice, no flakes), 45-check wiring suite, advisory-parallel + env-consistency gates, smoke test, shellcheck, ruff — all green.

PR #622, opened 2026-09-18
ci(codeql): add CodeQL python analysis workflow

itsmiso-ai:
Static security analysis gate for #621. CodeQL runs the explicit `security-and-quality` query suite over the checkout; the python arm covers `pr_reviewer/` plus the Python helpers under `scripts/` — the command-line-injection, path-traversal, SSRF and token-handling classes named in the SECURITY.md threat model that gitleaks/shellcheck/pytest can't see semantically. (The implicit v4 suite is `default`, so the suite is pinned explicitly rather than assumed.)

Bash is intentionally **not** in the matrix: CodeQL ships no bash extractor — the first revision of this PR proved it (init hard-fails with "Did not recognize the following languages: bash"). Shell scripts stay covered by the existing shellcheck gate in ci.yaml.

Scope note: an earlier revision also bundled a repo-wide ruff-format bootstrap (~105 files) that #621 never asked for; that has been dropped — bulk reformatting of security-sensitive modules deserves its own reviewed change, not a ride-along on a workflow addition.

Fixes #621

_Originally opened by foreman (workload wl-misospace-pr-reviewer-action-621); reworked by agentic-local after the bash matrix was found to be unsatisfiable._

PR #612, opened 2026-09-17
feat(review): add bounded specialist advisory-contract module

itsmiso-ai:
Closes #607. Pure deterministic contract layer — no orchestration, no model calls.

**Scope correction.** The previous head of this branch was a single commit that swept ~123 files (a repo-wide refactor plus a new `ruff.toml`) in alongside the specialists work. That was out of scope for #607 and parked as `needs-human` because the diff was unreviewable. Rebuilt onto `main` with only what the ticket asks for:

- `pr_reviewer/specialists.py` — parser/normalizer + fence-safe Markdown renderer
- `scripts/prompt_fragments/specialist_{correctness,security,tests}.txt`
- `tests/test_specialists.py` (79 cases)
- one AGENTS.md bullet

**The contract**
- Three **fixed** roles — `correctness`, `security`, `tests` — a closed `SPECIALIST_ROLES` set with no custom-role surface.
- One shared **version-1** advisory contract: `role` / `leads` / `truncated` / `truncation` / `errors`.
- **Fail-soft:** malformed JSON, unknown role, non-object payload, non-array `leads`, or unusable entries degrade to a populated `errors` list with empty/partial leads — never an exception, never fabricated leads.
- **Severity capped below a blocker:** only `major`/`minor`/`info`; `blocker`/`critical` aliases downgrade to `major`. A specialist lead can never set `has_blocker` or flip the verdict — enforcement stays the final reviewer's job.
- **Bounded & deterministic:** `MAX_LEADS` / `MAX_MESSAGE_CHARS` caps, declared lead order, first-wins dedup, and a hard UTF-8 byte cap in the renderer that drops whole trailing lines and never splits a multibyte character.
- **Trust-framed role prompts:** supplied repo/PR text is untrusted data (not instructions), strict JSON only, and advisory leads that require final verification — not an approve/request-changes.

**Non-goals honored:** no model/network calls, no execution of lead content, no corpus/publish wiring, no per-role provider config, no custom prompts, no parallelism/timeouts.

**Checks:** CI `validate` — py_compile, shellcheck, all bash suites, gitleaks, actionlint, pytest (coverage 82%, gate 72). Contract reviewed against every #607 acceptance criterion.

PR #630, opened 2026-09-20
feat(evals): specialist-mode eval fixtures and effectiveness telemetry (#610)

### LINKED ISSUE #610: [P2] Deep review: add specialist-mode eval fixtures and effectiveness telemetry
Parent: #580
Depends on: #609

## Summary

Extend the existing eval harness so deep specialist mode can be measured against the standard single-reviewer path before considering any default change or more expensive specialist tool loops.

The question is not merely whether three specialist calls run. The evals must show whether advisory specialist leads improve final-review capability **without** creating indiscriminate findings, duplicate noise, runaway tool use, or unacceptable latency/cost.

Reuse the repository-context eval framework from #577 rather than creating a second benchmark system.

## Required scenarios

Add at least four specialist-focused fixtures.

### 1. Security-specialist value

A PR where the ordinary diff looks plausible but a trust-boundary/auth/path/injection issue is easy to miss unless explicitly considered from a security perspective.

Expected behavior in deep mode:

- security specialist raises a relevant lead;
- final reviewer verifies it;
- final finding cites/mentions the concrete evidence, not merely the specialist assertion.

### 2. Correctness/regression-specialist value

A cross-file/state/compatibility regression where correctness-oriented reasoning should generate a useful lead that the final reviewer can verify with existing repository context/tools.

### 3. Test-specialist value

A change whose implementation bug or compatibility risk is exposed by an existing test/spec/migration pattern. The tests specialist should point the final reviewer toward the gap without being told the exact file/path in the fixture prompt.

### 4. Negative control: clean PR

A clean change where specialists may legitimately return no leads.

Expected behavior:

- final review remains clean;
- deep mode does not invent findings just because three extra model passes ran;
- duplicate/generic style commentary is penalized;
- native-tool usage stays within a conservative threshold.

Optional fifth scenario: overlapping specialist leads where correctness + security identify the same root issue and the final reviewer should publish one deduplicated finding.

## A/B matrix

For these fixtures compare at least:

- standard mode (`deep_review=false`);
- deep mode (`deep_review=true`).

Where the harness already supports repeated runs, run enough repetitions to expose model variance rather than treating one pass as a benchmark.

Do not require live external services in normal CI.

## Scoring / telemetry

Extend result metadata only as needed to capture generic signals such as:

- specialist role success/error;
- lead count per role;
- whether an expected lead category appeared;
- whether the final review verified/adopted the expected issue;
- duplicate lead/finding count;
- final native tool-call count;
- total specialist wall-clock duration;
- standard vs deep final verdict/finding capability pass;
- false-positive/negative-control failure.

Prefer capability checks over brittle prose matching.

If possible, distinguish:

- `lead_generated`
- `lead_verified_in_final_review`
- `lead_rejected_or_unused`

This makes it possible to tell whether a specialist is useful even when the final reviewer correctly discards some suggestions.

## Success criteria

Do **not** encode a universal model-independent benchmark score in production behavior.

For this ticket, success means the harness can demonstrate all of the following on deterministic synthetic traces and can measure them on real model runs:

- deep mode can improve at least one specialist-targeted capability case;
- final reviewer can reject an intentionally bad/unsupported specialist lead;
- overlapping leads can deduplicate into one final finding;
- clean negative control does not require findings or excessive tool calls;
- specialist failures are visible in telemetry without invalidating an otherwise valid final review.

## Expected files

Likely:

- `evals/corpus-specialists.json` or equivalent separate corpus
- `scripts/eval_harness.py`
- `.github/workflows/eval-harness.yaml`
- scorer/schema tests
- README/AGENTS eval documentation

Prefer a separate specialist corpus so #577's repository-context benchmark remains stable.

## Required tests

- specialist fixture schema validation;
- synthetic trace where expected specialist lead passes;
- synthetic trace where absent lead fails;
- synthetic unsupported lead is correctly rejected by final-review scoring;
- duplicate specialist leads do not satisfy a dedupe check unless final output is singular;
- clean negative control fails when fabricated findings/tool spam exceed its bound;
- standard-mode scoring remains unchanged;
- workflow matrix can select standard vs deep and repeated runs;
- telemetry parsing tolerates one or more failed specialist roles.

## Acceptance criteria

- [ ] At least four specialist-mode capability fixtures exist.
- [ ] Standard vs deep A/B execution is supported.
- [ ] Scoring distinguishes advisory lead generation from final verified findings.
- [ ] Negative control penalizes hallucinated/duplicate findings and unnecessary exploration.
- [ ] Existing #577 repo-context evals remain behaviorally unchanged.
- [ ] Repeated-run specialist evaluation is possible for Qwen3.8-27B-class and future models.
- [ ] No production default is changed solely by this ticket.

## Non-goals

- Enabling deep review by default.
- Per-role model/provider routing.
- Specialist native tool loops.
- Establishing a universal quality threshold.
- Persistent semantic indexing.
- Historical-rule mining.

itsmiso-ai:
## Summary

Implements #610: extends the existing repository-context eval framework (no second benchmark system) so the deep-review specialist phase can be A/B-measured against the standard single-reviewer path, with capability-based scoring on deterministic synthetic traces and real-model measurement plumbing.

- **New corpus `evals/corpus-specialists.json`** — 5 fixtures over real merged PRs, each with grader-only `specialist_expectations` split into two grading scopes:
  - `lead_checks` (deep-only diagnostics, graded only on `<mode>+deep` runs): `lead_generated`, `lead_disposition` — and `verified` demands concrete evidence (mandatory `finding_file_any`; a finding that merely repeats the specialist's wording with no file grounding computes `rejected`, never `verified`).
  - `effectiveness_checks` (the comparable A/B subset, graded on standard **and** deep): `final_findings_count` / `dedupe_final_findings` over the production finding shape, so the headline measures the same final-review capability on both sides.
  - Fixtures: #547 security value (JWT/OIDC cache race) + hallucination guard (fabricated secret-leak must not be adopted); #551 correctness value (carry-forward `needs_full_review` loop); #612 tests value (severity-alias/cap normalization coverage); #622 negative control (clean CI PR: no findings, no duplicates, `max_tool_calls` ≤ 4); #597 overlap/dedupe (security+correctness leads → ≤ 1 grounded final finding).
- **Harness (`scripts/eval_harness.py`)**:
  - `--deep-review false|true|both`; deep runs labelled `<mode>+deep`, executed across the same `--runs-per-mode` repetition (A/B matrix; dry-run lists planned labels).
  - Real execution boundary: passes `REPO` + `PR_NUMBER` into `run_review.sh` (the pipeline hard-requires both), consumes the validated `ai-output.json` artifact (verdict, markdown, production-shape findings, `verdict_source`) — not `verdict.json`, which the pipeline never writes — with model from `analysis_engine.txt` and tokens from the per-tier `ai-response.*.json` usage; stale per-run artifacts (incl. symlinked ones) reset before each run.
  - Fail-soft `load_specialist_telemetry()` tolerant of failed/missing/malformed/derived role artifacts.
  - Report: per-mode `specialist_effectiveness_runs/_passes/_pass_rate` (comparable on both labels) and `specialist_lead_runs/_passes/_pass_rate` (deep only); per-run scoped check detail; per-PR rate dicts.
- **Workflow `eval-harness.yaml`**: `deep` dispatch input (`false|true|both`, default `false`) → `--deep-review` — added in commit ecdbc6e (not re-shown in the 70a8984 incremental delta; see that commit's diff).
- **Docs**: AGENTS.md runbook + README "Specialist corpus & deep A/B" sections updated to the two-scope semantics.

## Explicit non-changes (per issue non-goals)

- No production default changed: `deep_review` remains off by default; `action.yml` untouched.
- #577 repo-context evals untouched behaviorally (existing suites green, `evaluate_capability` unchanged).
- No per-role routing, no specialist tool loops, no universal quality threshold encoded in production behavior.

## Tests (all offline, no live services in normal CI)

- `tests/test_eval_harness_specialists.py`: scorer semantics both directions; production finding shape (`message`/`file`/`line`); parroting-not-verified; errored runs; unknown types fail soft; telemetry parsing incl. failed roles.
- `tests/test_eval_harness_specialists_corpus.py`: two-list schema + closed-set validation (verified-dispositions-must-ground, negative-control shape, dedupe-both-roles, comparability rules), and deterministic synthetic-trace grading in both directions incl. standard-run effectiveness passes/fails and parroting deep runs.
- `tests/test_eval_harness_boundary.py`: real subprocess boundary via a fake orchestrator — env contract snapshot (`REPO`/`PR_NUMBER`/`DEEP_REVIEW`/`TOOL_MODE`), stale-artifact reset (incl. symlinked stale artifact unlinked-as-link), `ai-output.json` consumption, stdout fallback, malformed/symlinked specialist aggregate derived fallback, control-byte tolerance, and predicates never touching the filesystem (adversarial path-shaped inputs; per the repo's #252 convention).
- Locally: `2443 passed, 1 skipped`; coverage gate (`--cov-fail-under=72`) at 83.5%. The only two failing tests (`test_evidence_providers.py::TestMarkdownCaps`) fail identically on pristine `main` in this sandbox (no `python3` binary for spawned providers) and are environment-only.

## A/B + telemetry demo

`python scripts/eval_harness.py --corpus evals/corpus-specialists.json --modes native_loop --deep-review both --runs-per-mode 2 --dry-run` plans standard + `<mode>+deep` runs for every fixture. Real-model measurement: Actions → Eval harness → corpus `evals/corpus-specialists.json`, deep `both`.

Closes #610


joryirving (changes requested):
Requesting changes on exact head `ecdbc6e47e3e98097074f27666ccd98ba95ca681`.

The #610 fixture/scorer structure is promising, but three acceptance-level gaps remain:

1. **The advertised real-model execution path is not actually wired to `run_review.sh`.** `run_review_for_pr()` never sets `REPO` or `PR_NUMBER`, while `scripts/sections/config.sh` exits when either is missing. A workflow_dispatch eval therefore cannot select the corpus PR. Even after that is fixed, the harness looks for `verdict.json`, but the review pipeline's validated artifact is `ai-output.json`. As written, real runs either fail before review or fall back to stdout instead of loading the final review. Please set the repo/PR env from the corpus entry, consume the actual `ai-output.json` artifact, and add a test that exercises this integration boundary rather than only synthetic `ReviewRun` objects.

2. **Real final findings cannot satisfy the new specialist scorer.** `run_review_for_pr()` never populates `ReviewRun.findings`; additionally, the production `ai-output.json` finding schema uses `message` / `file` / `line`, while `_finding_predicate_matches()` and the new synthetic tests use `description`. That makes positive `lead_disposition=verified`, `final_findings_count`, and dedupe checks fail on real output, while `not_adopted` can falsely pass because the matcher sees zero matching findings. Normalize the production finding artifact into one scorer shape (or make the scorer understand the production shape) and cover it with a real ai-output-shaped fixture.

3. **The A/B headline is partly self-fulfilling and `verified` does not prove verification.** Every positive specialist fixture contains `lead_generated`/lead-based checks, so a standard run (`deep_review=false`) necessarily fails `specialist_capability` because it cannot emit specialist leads. A deep run can therefore "improve" this metric by construction even if the final review is no better. Separately, `lead_disposition=verified` currently means only "a final finding matched loose category/message substrings"; it does not require the concrete evidence/file/tool verification that #610 explicitly calls for. Please keep deep-only lead-generation telemetry separate from the **comparable** standard-vs-deep final-review capability score, and make the positive adoption check require concrete verification evidence (for example a matching file/line/evidence predicate or an existing capability-style evidence check).

CI and CodeQL are green on this head, and the five-fixture corpus / fail-soft specialist telemetry direction looks good. These are focused measurement-correctness issues rather than a request to redesign deep review.

itsmiso-ai (commented):
Re-review requested: the three REQUEST_CHANGES blockers are addressed in 70a8984 (execution boundary REPO/PR_NUMBER + ai-output.json consumption with subprocess-level boundary tests; production five-key finding shape through all finding predicates; lead_checks deep-only vs effectiveness_checks comparable-A/B, with verified now demanding concrete file grounding). The reviewer's note that the workflow `deep` input was not visible in the incremental delta: it landed in commit ecdbc6e (see that commit's .github/workflows/eval-harness.yaml diff) — 70a8984 intentionally did not touch the workflow. The edge-case-path must-check gap (null bytes/symlinks) is addressed in the follow-up commit with adversarial boundary tests feeding the boundary tokens themselves per AGENTS.md (#252 convention). Adding the ai-review label to force a fresh full-scope review.

joryirving (changes requested):
Requesting changes on exact head `28fc339b03f03dc1a181ea785111a4a499e67e44`.

The three blockers from my previous review are fixed: the harness now sets `REPO`/`PR_NUMBER`, consumes `ai-output.json` and production-shape findings, and separates deep-only lead diagnostics from the comparable standard-vs-deep effectiveness score. CI and CodeQL are green on this head.

One real execution-boundary blocker remains:

**Override `GITHUB_WORKSPACE` for the review subprocess to the harness clone.** `run_review_for_pr()` starts from `env = os.environ.copy()` and runs `run_review.sh` with `cwd=repo_path`, but it never replaces an inherited `GITHUB_WORKSPACE`. In GitHub Actions, `GITHUB_WORKSPACE` points at the workflow checkout, not the harness's temporary clone. Several production review paths explicitly prefer that env var over cwd, including repo-map generation, related-code context, specialist artifact roots, SARIF/evidence roots, carry-forward/enforcement workspace roots, and other workspace-guarded helpers. In a real eval run this can therefore produce a split-brain review: PR/diff metadata comes from the corpus entry while filesystem context and specialist artifacts come from or are written into the workflow checkout. Deep mode is especially affected because `run_specialists.py` defaults its artifact root to `$GITHUB_WORKSPACE`, while the eval harness later reads specialist telemetry from `repo_path`.

Please set `env["GITHUB_WORKSPACE"] = str(repo_path)` before invoking the orchestrator (or equivalently remove the inherited value only if every downstream helper is guaranteed to fall back to cwd; explicitly setting it is safer and matches production semantics). Add a boundary regression test that seeds the parent process with a deliberately different `GITHUB_WORKSPACE` and verifies the fake orchestrator sees the temp clone path, with artifacts/context staying inside that clone.

This is narrowly scoped. I do not see another blocker in the #610 scoring/telemetry changes on this head.

itsmiso-ai:

**Total attempts:** 126


**Attempts by lane:**
- **NEEDS_HUMAN:** 1 attempt
- **NORMAL:** 125 attempts


**Failing run(s):**
- https://github.com/misospace/pr-reviewer-action/security/code-scanning/106
- https://litellm.jory.dev/v1
- https://github.com/misospace/pr-reviewer-action/security/code-scanning/107
- https://github.com/misospace/pr-reviewer-action/security/code-scanning/108

**Last attempt:** Requesting changes on exact head `28fc339b03f03dc1a181ea785111a4a499e67e44`.

The three blockers from my previous review are fixed: the harness now sets `REPO`/`PR_NUMBER`, consumes `ai-output.json` and production-shape findings, and separates deep-only lead diagnostics from the comparable standard-vs-deep effectiveness score. CI and CodeQL are green on this head.

One real execution-boundary blocker remains:

**Override `GITHUB_WORKSPACE` for the review subprocess to the harness clone.** `run_review_for_pr()` starts from `env = os.environ.copy()` and runs `run_review.sh` with `cwd=repo_path`, but it never replaces an inherited `GITHUB_WORKSPACE`. In GitHub Actions, `GITHUB_WORKSPACE` points at the workflow checkout, not the harness's temporary clone. Several production review paths explicitly prefer that env var over cwd, including repo-map generation, related-code context, specialist artifact roots, SARIF/evidence roots, carry-forward/enforcement workspace roots, and other workspace-guarded helpers. In a real eval run this can therefore produce a split-brain review: PR/diff metadata comes from the corpus entry while filesystem context and specialist artifacts come from or are written into the workflow checkout. Deep mode is especially affected because `run_specialists.py` defaults its artifact root to `$GITHUB_WORKSPACE`, while the eval harness later reads specialist telemetry from `repo_path`.

Please set `env["GITHUB_WORKSPACE"] = str(repo_path)` before invoking the orchestrator (or equivalently remove the inherited value only if every downstream helper is guaranteed to fall back to cwd; explicitly setting it is safer and matches production semantics). Add a boundary regression test that seeds the parent process with a deliberately different `GITHUB_WORKSPACE` and verifies the fake orchestrator sees the temp clone path, with artifacts/context staying inside that clone.

This is narrowly scoped. I do not see another blocker in the #610 scoring/telemetry changes on this head.

Posted automatically by Dispatch on 2026-09-20T13:36:08.730Z

joryirving (changes requested):
Requesting changes on exact head `e6ce9ecfac05ff4cf1716d0aacbf29cce20780bb`.

The `GITHUB_WORKSPACE` pin is fixed correctly and the new boundary tests cover the ambient-Actions-workspace failure mode. CI and CodeQL are green on this head.

One remaining real-model correctness issue surfaced from tracing the full historical-fixture path:

**The eval worktree is never checked out to the corpus PR's head revision.** `run_review_for_pr()` clones `https://github.com/<repo>.git` once and then reuses that checkout for every corpus entry, but there is no fetch/checkout/reset to PR `#<number>`'s head before `run_review.sh` executes. `run_review.sh` does fetch the requested PR object/diff from the API, so the result is a mixed-time review: the PR metadata/diff is historical, while repo-map generation, related-code context, manifest reads, native `read_file`/`git_grep`/tree exploration, and specialist verification operate on whatever revision the clone currently has (normally today's default branch).

That is especially material for this #610 corpus because the fixtures are historical merged PRs (#547/#551/#612/#622/#597). Production usage checks out the PR head before review; the eval harness needs to do the same or it is not measuring the same capability.

Please, before each run, deterministically fetch and checkout the corpus PR head (for this GitHub-specific harness, e.g. resolve/fetch `refs/pull/<PR_NUMBER>/head` or resolve the PR head SHA through the API and checkout that SHA). The checkout must happen for every PR entry because the same `repo_path` is reused across fixtures/modes/repetitions. Fail the run clearly if the requested PR revision cannot be materialized rather than silently evaluating against default-branch HEAD.

Add a regression test that creates or mocks two distinct PR revisions and proves consecutive corpus runs observe the matching revision in the worktree, not the initial/default branch. Ideally also assert the fake orchestrator sees `git rev-parse HEAD` equal to the expected PR head.

I do not see another blocker in the workspace pin or the specialist scoring changes. This is the last execution-fidelity gap I found while following a real historical A/B run end to end.

joryirving (approved):
Approved on exact head `0025d328b2d4e999058be4921c0f9e07748e0dbe`.

The historical-fixture execution path is now coherent end to end:

- every run fetches `refs/pull/<PR>/head` and force-detaches to that exact revision before context generation;
- the checked-out SHA is recorded on the run;
- reused worktrees are re-materialized for every fixture/mode/repetition;
- missing PR refs fail closed instead of silently evaluating default-branch HEAD;
- `GITHUB_WORKSPACE` remains pinned to the temp clone;
- the regression suite exercises `547 -> 551 -> 547` in one reused checkout and verifies the orchestrator subprocess itself sees the expected `git rev-parse HEAD`.

The earlier REPO/PR_NUMBER, ai-output.json, production finding-shape, A/B metric separation, concrete verification grounding, and workspace-root issues remain resolved. CI and CodeQL are green on this head. I do not have another blocker for #610.

